PR #3, opened 2018-05-18
Preliminary Russian Translation

dikonov:
I cannot properly test because I do not have the supported watch


piggz:
Thank you!  Do you plan on getting the watch?

PR #18, opened 2018-06-05
Update harbour-amazfish-sv.ts

eson57:
new strings translated


eson57:
There's something wrong with your repo, it is producing conflicts on it's own where there should not be any. This is the second time it happens.

piggz:
Odd .... not had time to look into that, but thx

PR #30, opened 2018-07-09
Update harbour-amazfish-sv.ts

eson57:


eson57:
new strings translated

PR #64, opened 2019-03-29
Refreshed app icon & dutch translation

JSEHV:
Refreshed app icon


piggz:
Could you split out the translation and icon into separate PRs?  Iike the translation now, but some changes to the icon.

PR #69, opened 2019-04-05
Refreshed app icon

JSEHV:
Refreshed app icon


piggz:
how about making it more watch-like instead of clock-like? and then adding a heart-beat across the middle, to indicate health monitoring?

PR #84, opened 2019-09-09
Improved sleep analysis

vigejolla:


vigejolla:
At least with firmware 1.1.5.16 the original code resulted in absolutely insane results, claiming that I slept most of the day. With these changes the values are at least quite near sane.

piggz:
Could you re-base this and i'll give it a serious look.

vigejolla:
> Could you re-base this and i'll give it a serious look.

I could, but I doubt it makes sense to merge it even after that. But giving that serious look is probably a good idea. What I mean is that it doesn't work like I would like to work even after my changes, even though it is a lot better. So after that serious look you might have the idea about what should really be done... In any case, I'll rebase it, maybe in the evening today.

piggz:
I have merged some improved sleep analysis

vigejolla:
> I have merged some improved sleep analysis

That sounds awesome, I guess I'll have to test it asap :)

PR #89, opened 2019-09-18
Fix right wrist assignment in mibandservice.cpp

threepistons:
Whilst perusing the code to try to figure out why my Band 2's heart rate data is no longer appearing in the app, I spotted that the switch allowing the user to specify which wrist they wear their watch on specified the left wrist twice.  So I fixed it, assuming that I read the keys in the correct order.  The app builds and runs on my Xperia XA2 running SFOS 3.1.0.11.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+harbour-amazfish.pro.user
--- a/daemon/src/services/mibandservice.cpp
+++ b/daemon/src/services/mibandservice.cpp
@@ -360,7 +360,7 @@ void MiBandService::setWearLocation()
         writeValue(UUID_CHARACTERISTIC_MIBAND_USER_SETTINGS, QByteArray(WEAR_LOCATION_LEFT_WRIST, 4));
         break;
     case 1:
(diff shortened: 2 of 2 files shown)


piggz:
does it sync the data and not show it, or does sync fail? one user had a problem with a corrupt database for example.

another issue is that the last sync time can become wrong...

do
dconf read /uk/co/piggz/amazfish/device/lastactivitysyncmillis

if you convert that to seconds, then a time since epoch, you will find out when it is syncing from

if that doesnt change after doing a sync in app, then add an hour or so to it, 1 hr = 3600000

the problem occurs when timezone changes, eg, going on holiday.  i need to investigate a proper fix by seeing how gadgetbride sends time to the watch.

threepistons:
```
[nemo@Sailfish harbour-amazfish]$ dconf read /uk/co/piggz/amazfish/device/lastactivitysyncmillis
int64 1564787820000
```
Wed, 18 Sep 2019 10:48:53 GMT according to the epoch converter I found.

Yet,
```
sqlite> select max(timestamp_dt) from mi_band_activity ;
max(timestamp_dt)
2019-08-02T23:16:00
```

PR #104, opened 2020-02-27
little bug-fix for time format selection 

AlphaX2:
the time format selection was swapped, now the watch follows the gui (bug was reported by erdzas on OpenRepos comments)


piggz:
Thanks confirmed against mibandservice.cpp

PR #106, opened 2020-03-01
Updated GUI and changes to the settings pages.

AlphaX2:
Hey, 

I've tried to make some changes to get a more pleasing GUI here and there. Start page got a Circle element to show steps, battery is presented by a bar. Settings menu is reworked to look more like the SailfishOS settings, also the different settings pages are using SectionHeader elements now. Hope you like it, add whatever you like to. :-)


piggz:
Looks great.  Would it be much work to split this into separate PRs for each function?

AlphaX2:
Should be possible, but needs some work, because it's also in my forked repo one bigger commit.

piggz:
If you reset to head^1, then create  a branch for the settings page, then add the files as required.  I could potentially try them out in an hour or so.

PR #108, opened 2020-03-02
reworked settings menu, with Icon and ListItem, to imitate the Sailfi…

AlphaX2:
…shOS settings page


piggz:
Looks good, could you add some padding to the LHS though as the icons are a little close to the edge.  thx

piggz:
I think the model needs a way to hide/disable items based on the supported features.  FEATURE_ALARMS and FEATURE_WEATHER are the relevant ones.

PR #116, opened 2020-03-15
UI refactoring

mentaljam:
I've done some UI improvements (as it seems to me). If you find them useful, please merge this PR.


piggz:
Looks good, will hopefully test tomorrow.  Thanks.

piggz:
Testing atm

ui/qml/pages/UnpairDeviceDialog.qml
```diff
@@ -0,0 +1,46 @@
...
+
+Dialog {
+    acceptDestination: Qt.resolvedUrl("PairSelectDeviceType.qml")
+    acceptDestinationAction: PageStackAction.Replace
+
+    onAccepted: {
+        DaemonInterfaceInstance.disconnect()
+        AmazfishConfig.pairedAddress = ""
+        AmazfishConfig.pairedName = ""
+        AmazfishConfig.deviceAuthKey = ""
```
piggz: Im not sure about wiping the auth key....if a user decides to re-pair (like me while testing!) they will have to enter the whole key again which is a pain.  thoughts?

mentaljam: Possible solution is in the last commit

piggz:
Just tried the latest version and the pairing key was empty when I went to pair again.

piggz:
Also, pairing doesnt work.  When clicking on the device in the scan result list, your get
Pairpage.qml:64 ....AmazfishConfig is not defined

piggz:
Sorry, retract that, i must have been using an old build.....just tried again and didnt get that error.  However, i was still prompted to enter an auth key, when I did, i just entered random chars as i just wanted to see if the error happened, and low and behold, it paired ok, even though id entered the wrong key.

mentaljam:
Sorry, I don't have a device that requires an auth key. I just tried to enter a random string to the field and accept the dialog. After that I returned back and opened the auth dialog again. The previously entered random key was there.

About a success pairing with a random key - I have no idea. To my shame I don't understand how does pairing work.

piggz:
After some more testing, it seems to work.  Pairing seems to work better if the device is removed from the sailfish BT settings dialog. Please make future PRs a little smaller, but thanks for al this work :)

PR #147, opened 2020-12-19
Add WIP clickable.json

jonnius:
This app seems to make use of the [Convergence Components](https://gitlab.com/dylanvanassche/convergence-components) and therefore theoretically support Ubuntu Touch (UT). I was curious to see if we can build an Ubuntu Touch click package with [Clickable](https://clickable-ut.dev/en/latest/) (even though I don't have the smart watch), so I did the first step, adding a `clickable.json`.

I ran into `Project ERROR: mpris-qt5 development package not found`. I assume the package needed in Ubuntu is `libmpris-qt5-dev`, which is in Ubuntu 20.04, but not 16.04, which Ubuntu Touch is still based on.

So I got a few questions:
- What is the minimum Qt version required? (UT has 5.9, but is shifting to 5.12 soon)
- What other dependencies are there besides `mpris-qt5`? (maybe you could add build instructions to the `README.md`?)
- Do you think it is fairly easy to build missing dependencies like `mpris-qt5` so we can ship them with the app?
- Any special system service or other integration needed?

Even if it might not be possible at the moment, this can serve as a starting point for whoever wants to pick it up later, maybe after Ubuntu Touch shifted to Ubuntu 20.04.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,6 @@
 harbour-amazfish.pro.user
+
+# Clickable
+/.clickable/
+/build/
+/libs/
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -1,0 +1,26 @@
+# Build
+
+## Ubuntu Touch
+
+[Clickable](https://clickable-ut.dev/en/latest/install.html) is used to build and
+debug the app for Ubuntu Touch.
+
+Download dependencies:
+
+```bash
+clickable prepare-deps
+```
+
+Build dependencies:
+
+```bash
+clickable build-libs --arch arm64 # or amd64 or armhf
+```
+
+Build the app:
+
+```bash
+clickable build --arch arm64 # or amd64 or armhf
+```
+
+For more details see [Clickable Documentation](https://clickable-ut.dev/en/latest/usage.html).
--- a/clickable.json
+++ b/clickable.json
@@ -3,5 +3,16 @@
   "builder": "qmake",
   "build_args": [
     "FLAVOR=uuitk"
-  ]
+  ],
+  "env_vars": {
+    "PKG_CONFIG_PATH": "${QTMPRIS_LIB_INSTALL_DIR}/usr/lib/${ARCH_TRIPLET}/pkgconfig"
+  },
+  "scripts": {
+    "prepare-deps": "git clone https://git.sailfishos.org/mer-core/qtmpris.git libs/qtmpris"
+  },
+  "libraries": {
+    "qtmpris": {
+      "builder": "qmake"
+    }
+  }
 }
--- a/clickable.json
+++ b/clickable.json
@@ -1,0 +1,7 @@
+{
+  "clickable_minimum_required": "6.20.0",
+  "builder": "qmake",
+  "build_args": [
+    "FLAVOR=uuitk"
+  ]
+}


jonnius:
Just saw you directly reference `platform.silica` in [ui/ui.pro](https://github.com/piggz/harbour-amazfish/blob/master/ui/ui.pro#L45). Won't that break compatibility with other platforms?

piggz:
It does ... i have a few changes locally while the convergence components work to become more platform independent ... hopefully in the future we can just include as a submodule.

There are probably several silica assumptions, but the plan is to move toward platform abstraction, so any PRs welcome.

Min qt version is 5.6 for sailfish.  Perhaps a good place to start for deps is pure-maps, which already support UT, and has a lot of similar deps.

dano6:
https://gitlab.manjaro.org/manjaro-arm/packages/community/plasma-mobile/amazfish/-/blob/master/PKGBUILD
Build file for manjaro-arm KDE for reference

piggz:
Wow awesome ...
There is one other dep that is a pain to build, mapbox-gl-native and its bindings.  Rinigus has a fork which includes installation steps as upstream doesnt have that, however, rinigus version doesnt build on my laptop due to too new boost.  The qml components do build ok though.  These are a runtime dep for the UI to display maps when you have synced back a GPS activity.

jonnius:
@dano6, thanks for the link. 

Btw, I am the one who did the Clickable build configuration of Pure Maps and OSM Scout Server for Ubuntu Touch based on what was already there for Sailfish, while rinigus implemented the QML platfrom support. For Pure Maps we have a [Clickable build configuration](https://github.com/rinigus/pure-maps/blob/master/packaging/click/full-build.json) including building `mapbox-gl-native`.

So maybe we only need to build the `mpris-qt5` for Ubuntu 16.04. Is this thing part of official Qt? Because I didn't find it in the Qt documentation.

dano6:
No it is just Sailfish thing. It isn't even packaged for Arch. I has to install it from source to run it on my Arch setup.

piggz:
For opensuse, it is packaged as part of the deepin music player ... but probably better to sue upstream version for bug fixes

jonnius:
Could someone provide a link to the upstream project of this `mpris-qt5` thingie?

piggz:
https://git.sailfishos.org/mer-core/qtmpris

jonnius:
Thanks! Compiling qtmpris was pretty straight forward. I added it to the clickable config and provided an `INSTALL.md` with instructions.

Next dependencies I am looking into at the moment:
```
Project ERROR: Unknown module(s) in QT: KDb3 KContacts
```

jonnius:
I tried building KContacts from [Sources](https://invent.kde.org/frameworks/kcontacts), but it requires Qt 5.14, while we only have Qt 5.9 in UT. Should I build some older version of KContacts or how can I circumvent that issue?

jonnius:
Seems like KContacts `v18.11.80` is the last version with Qt 5.9 support. But building it requires other Qt stuff in versions not available in Ubuntu 16.04. Is there some shortcut or do I need to go down this road an compile everything and ship with the app?

piggz:
I dont actually use kcontacts yet, you could just comment it out of the watchfish pri file for now i think

jonnius:
Ok, will do. And what about `KDb3`? What is that? Do I really need it? And where should I get sources from?

piggz:
Yes, you most certainly do ... its one of the core parts of amazfish.  KDB is the database library used by Kexi.  Amazfish stores all healt/activity info in a Kexi database, which allows the user to analyse their own data within Kexi if they wish.  It was one of my ideas to empower users with their own data, instead of having it all in the cloud.  Whether or not people make use of that is another matter!!

piggz:
Its in the kde repos, eg https://github.com/KDE/kdb

dano6:
Sadly, it is not in xenial: https://launchpad.net/ubuntu/+source/kdb

jonnius:
I got kdb built, but the app won't find it (`Project ERROR: Unknown module(s) in QT: KDb3`). Is there some env var I can set to tell it where to look for it? I tried providing `PKG_CONFIG_PATH` to the pkgconfig directory, but that didn't help.

dano6:
`QML_IMPORT_PATH += path/to/module`
path should be to kdb3 folder where are .so files

jonnius:
I tried that without success. E.g. for arm I do `QML_IMPORT_PATH+=<kdb-install-prefix>/lib/arm-linux-gnueabihf`.

This is how the kdb install prefix looks like:
```
├── bin
│   └── kdb3_sqlite3_dump
├── include
│   └── KDb3
│       ├── config-kdb.h
│       ├── KDb
│       ├── [... and a lot more]
├── lib
│   └── arm-linux-gnueabihf
│       ├── cmake
│       │   └── KDb3
│       │       ├── KDbConfig.cmake
│       │       ├── KDbConfigVersion.cmake
│       │       ├── KDbQCHTargets.cmake
│       │       ├── KDbTargets.cmake
│       │       └── KDbTargets-release.cmake
│       ├── libKDb3.so -> libKDb3.so.4
│       ├── libKDb3.so.4 -> libKDb3.so.4.3.0
│       ├── libKDb3.so.4.3.0
│       ├── pkgconfig
│       │   └── KDb3.pc
│       └── plugins
│           └── kdb3
│               ├── kdb_sqlitedriver.so
│               └── sqlite3
│                   └── kdb_sqlite_icu.so
└── mkspecs
    └── modules
        └── qt_KDb3.pri
```
I wonder whether I need to configure some path to `mkspecs/modules/qt_KDb3.pri`?

piggz:
I think you hit this bug
https://invent.kde.org/libraries/kdb/-/merge_requests/6
Perhaps if I switch detection to use pkgconfig it will work, or, you can try some of the tips in that bug

dano6:
Try just to copy mkspecs folder to lib/arm-linux-gnueabihf folder

jonnius:
I tried both without success.

### Change CMake includes order
As suggested I changed `kdb`'s CMakeLists.txt:
```diff
 # ECM
+include(KDEInstallDirs)
+include(KDECMakeSettings NO_POLICY_SCOPE)
+include(KDEFrameworkCompilerSettings NO_POLICY_SCOPE)
 include(ECMGeneratePriFile)
 include(ECMInstallIcons)
 include(ECMOptionalAddSubdirectory)
 include(ECMPoQmTools)
 include(ECMSetupVersion)
-include(KDEInstallDirs)
-include(KDECMakeSettings NO_POLICY_SCOPE)
-include(KDEFrameworkCompilerSettings NO_POLICY_SCOPE)
```
This didn't change anything about the structure of the installed files.

### Copy mkspecs
I copied the folder inside the kdb install prefix with `cp -r mkspecs/ lib/arm-linux-gnueabihf`. 

`QML_IMPORT_PATH` is set to `<kdb-install-prefix>/lib/arm-linux-gnueabihf`.

PR #150, opened 2020-12-21
Fix alarm page

carmenfdezb:
Hi @piggz!
I've noticed that alarm page doesn't work, with this commit it is fixed.
Regards.


piggz:
Sorry, you must try harder :)
I guess I missed this page in the port away from Silica.  Needs to use the non-silica components.

carmenfdezb:
Ok, sorry. I think #152 fix this without silica components. I close this PR

PR #164, opened 2021-02-09
Add some sfos specific notification icon definitions

jgibbon:
Another even more trivial one. 
Housekeeping, more or less.

cheers!


piggz:
Thanks, been meaning to add those.

PR #172, opened 2021-02-17
Fix resizing issue when using kirigami

dano6:
Idea taken from https://github.com/rinigus/pure-maps/blob/master/qml/README.md


piggz:
Thanks, can you add the required changes to the other pages too please?

piggz:
Thanks, manually added changes.

PR #183, opened 2021-05-03
rpm: Check if KCalendarCore installed using lua

Thaodan:
Don't assume os-release exists and use lua instead of check

Follow the link below for more information:
https://rpm.org/user_doc/lua.html


piggz:
Thanks, checked on new OBS and it builds

piggz:
Ah, not its not working ... i was deceived by an old build on obs which wasnt retriggered, but said success :/

PR #191, opened 2021-06-15
Fix Kirigami PageListPL

nicolasfella:
The code contains a few things that are unnecessary and actually break
things.

Now the page properly fills the width and the scrollbar is placed
properly

Before:
image

After:
image


piggz:
Thanks, keep it up ... the non-sailfish versions will be the least tested :)

PR #205, opened 2021-11-17
Simulate event support on devices which don't have it

vigejolla:


piggz:
This is a great feature addition, thankyou.

Im wondering if you could add some UI elements?  Perhaps a boolean to toggle on/off, and maybe a number to allow alert prior to event, maybe 0/15/30/45/60 minutes?

vigejolla:
> Im wondering if you could add some UI elements? Perhaps a boolean to toggle on/off, and maybe a number to allow alert prior to event, maybe 0/15/30/45/60 minutes?

On/Off setting is probably a good idea. I can add that, but I don't know when I'm going to have time to do that.

About alerting before the event: Wouldn't it be better if we fetched the alert time from the event? That is a bit more work, as it means we need to add the support to libwatchface as well, but I think it would be worth it. I can do that as well, but indeed I'm a bit low on free time :(

vigejolla:
Further testing has revealed that my QTimer based approach doesn't work in real life scenarios. I need to rewrite this using Nemo Keepalive.

PR #217, opened 2022-02-09
Fix Warnings regarding Connections in qml

marcoSchr:
This will fix all warnings regarding the "implicitly defined properties in Connections are deprecated" warnings

### DIFF
--- a/ui/qml/components/platform.generic/ExpandingSectionGroupPL.qml
+++ b/ui/qml/components/platform.generic/ExpandingSectionGroupPL.qml
@@ -71,7 +71,9 @@ Column {
 
             Connections {
                 target: section
-                onCloseAll: del.expanded = false
+                function onCloseAll() {
+                    del.expanded = false
+                }
             }
         }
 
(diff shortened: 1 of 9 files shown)


piggz:
Thanks for this, will test it later ... out of interest, what are you running amazfish on?

marcoSchr:
I am running on postmarketOS/Alpine Linux

piggz:
Bit of a problem, this syntax seems to have appeared in qt 5.14 ... but I need to support qt 5.6 on sailfish!

marcoSchr:
OK. Then I will need to check if there is a way to disable the warnings.

piggz:
There is, like this
https://github.com/rinigus/pure-maps/blob/master/src/main.cpp#L132

marcoSchr:
I saw you added this in bc9b9e377ae3908e91913bacbaa752f03f66ebd1.
I just tested it on postmarketOS, and I no longer get the warnings, thus I will close this pull request.

PR #221, opened 2022-03-27
Fix svg icon installation name

AndreySV:
Desktop file mentioned harbour-amazfish-ui icon and installed only
hardbour-amazfish svg icon, as a result after 'make install' no icon
was displayed because png icon are not rendered and svg icon is
installed under wrong name


piggz:
thx

PR #223, opened 2022-04-11
Reorganize build doc for dependency dependencies

TechnologyClassroom:
This is a slight reorganization of the compilation instructions which might help those that are not familiar with compiling software.

### DIFF
--- a/documentation/build-instructions.md
+++ b/documentation/build-instructions.md
@@ -31,20 +31,13 @@ etc as pointed out by build errors.
 
 ## In Ubuntu 20.04, with default Ubuntu repository, QT is in version 5.12.8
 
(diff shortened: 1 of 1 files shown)


piggz:
thx

TechnologyClassroom:
I plan to add build instructions for Mobian (PinePhone) and PureOS (Librem 5) once I get the kinks worked out too.  Should be the same instructions for both since they are essentially Debian Sid.  I got it to build and run with some additional dependencies [listed on this Pine forum thread](https://forum.pine64.org/showthread.php?tid=13122&pid=109029#pid109029).  I already had several QT packages installed so I want to try again fresh to make sure I have the correct list.

PR #236, opened 2022-09-03
Add support for the upload procedure of external resource for the PineTime (running InfiniTime).

JF002:
At InfiniTime, we are working on adding support for external resources : applications and watchfaces will be able to load images and fonts stored in the external SPI flash memory of the PineTime.

This is working quite well. See [here](https://github.com/InfiniTimeOrg/InfiniTime/issues/321) and [here](https://github.com/InfiniTimeOrg/InfiniTime/pull/1226).

Those resources are packaged into a ZIP file together with a "manifest file" that describes the mapping of the files into the watch. See [here](https://github.com/InfiniTimeOrg/InfiniTime/issues/1285).

Now, we need companion apps to implement the [upload procedure](https://github.com/InfiniTimeOrg/InfiniTime/issues/1285#issuecomment-1221605579) so users can easily update the resources from their device.

This PR contains the implementation of the Adafruit BLE File Transfer Protocol (the BLE FS API implemented in InfiniTime) and of the upload procedure specified for InfiniTime.

Note that the feature is not merged yet in InfiniTime, we are still doing some tests and waiting for other companion apps to add this feature.

Basically, this PR add `AdafruitBleFsService` which is the entry point for the Adafruit BLE File Transfer Protocol, `AdafruitBleFsService` which contains operations on the API (list, erase, write file) and `AdafruitBleFsWorker` which implements the upload procedure.

### DIFF
--- a/daemon/src/devices/infinitimefirmwareinfo.cpp
+++ b/daemon/src/devices/infinitimefirmwareinfo.cpp
@@ -44,7 +44,6 @@ void InfinitimeFirmwareInfo::determineFirmwareType()
                 qDebug() << "Resource file detected";
                 m_type = Res_Compressed;
             }
-
         }
     }
 }
--- a/daemon/src/devices/infinitimefirmwareinfo.cpp
+++ b/daemon/src/devices/infinitimefirmwareinfo.cpp
@@ -55,7 +55,10 @@ void InfinitimeFirmwareInfo::determineFirmwareVersion()
     case Firmware:
         m_version = "FW ()";
         break;
-    default:
+    case Res_Compressed:
         m_version = "Ressource ()";
+        break;
+    default:
+        m_version = "unknown";
     }
 }
--- a/daemon/src/devices/infinitimefirmwareinfo.cpp
+++ b/daemon/src/devices/infinitimefirmwareinfo.cpp
@@ -39,6 +39,12 @@ void InfinitimeFirmwareInfo::determineFirmwareType()
                 qDebug() << "DFU file detected";
                 m_type = Firmware;
             }
+            else if(root->entry("resources.json") != nullptr)
+            {
+                qDebug() << "Resource file detected";
+                m_type = Res_Compressed;
+            }
+
         }
     }
 }
@@ -50,6 +56,6 @@ void InfinitimeFirmwareInfo::determineFirmwareVersion()
         m_version = "FW ()";
         break;
     default:
-        m_version = "unknown";
+        m_version = "Ressource ()";
     }
 }
--- a/daemon/src/devices/pinetimejfdevice.cpp
+++ b/daemon/src/devices/pinetimejfdevice.cpp
@@ -10,8 +10,38 @@
 #include "hrmservice.h"
 #include "infinitimemotionservice.h"
 #include "infinitimeweatherservice.h"
-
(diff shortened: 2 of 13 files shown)


daemon/src/devices/infinitimefirmwareinfo.cpp
```diff
@@ -34,6 +39,6 @@ void InfinitimeFirmwareInfo::determineFirmwareVersion()
         m_version = "FW ()";
         break;
     default:
-        m_version = "unknown";
+        m_version = "Ressource ()";
```
piggz, line 42: Add case Res_Compressed here

daemon/src/devices/pinetimejfdevice.cpp
```diff
@@ -10,9 +10,39 @@
 #include "hrmservice.h"
 #include "infinitimemotionservice.h"
 #include "infinitimeweatherservice.h"
-
+#include "adafruitblefsservice.h"
 #include <QtXml/QtXml>
 
+namespace {
+    size_t GetMtuForCharacteristic(QString path, const char* characteristicUUID)
```
piggz, line 17: I think this should be moved to qble/qbledevice?

daemon/src/devices/infinitimefirmwareinfo.cpp
JF002: Done!

daemon/src/devices/pinetimejfdevice.cpp
JF002: I put this method in PinetimeJFDevice because it already contains some code that queries DBus to do the service discovery. Maybe we should move all this code to qble?

JF002: So, yes, you are right, this should be moved to qble/qbledevice, but I think this change will have impact in other device implementations that also do the service discovery. Is that OK if we do that in another PR?

JF002:
I've just rebased this branch on master, fixed conflicts and add a few steps in the upload procedure (create folder and subfolder, correctly remove older entries, remove obsolete files).

The [corresponding feature](https://github.com/InfiniTimeOrg/InfiniTime/pull/1299) has already been merged in InfiniTime, so this branch is ready for review :)

daemon/src/devices/pinetimejfdevice.cpp
piggz: ok

ui/qml/pages/PairSelectDeviceType.qml
```diff
@@ -15,7 +15,7 @@ PageListPL {
                     var pairpage = pageStack.push(Qt.resolvedUrl("./PairPage.qml"));
                     pairpage.deviceType = deviceType;})
             } else {
-                var pairpage = pageStack.push(Qt.resolvedUrl("./PairPage.qml"));
+                var pairpage = pageStack.push(Qt.resolvedUrl("/home/jf/git/harbour-amazfish/ui/qml/pages/PairPage.qml"));
```
piggz, line 18: I guess this shoudnt be here :)

JF002: Yes, sorry about that... I changed the path to debug an issue with the pairing page, and I shouldn't have committed that! I reverted that change.

PR #237, opened 2022-09-04
PineTime / InfiniTime : OTA now takes the DFU ZIP file

JF002:
In Infinitime, the firmware (for the OTA) is provided as a DFU ZIP file. However, Amazfish does not support the ZIP file and only accept the uncompressed binary file.

This PR adds support for the ZIP file (and removes the support for the raw bin file). The name of the firmware file inside the archive and the CRC of the firmware are read from the `manifest.json` file included in the DFU archive.

### DIFF
--- a/daemon/daemon.pro
+++ b/daemon/daemon.pro
@@ -14,7 +14,7 @@ TARGET = harbour-amazfishd
 
 LIBS += -Lqble/qble -L$$OUT_PWD/../lib -lamazfish -lz
 PKGCONFIG += dbus-1
-QT +=  positioning KDb3 network dbus
+QT +=  positioning KDb3 network dbus KArchive
 CONFIG += c++14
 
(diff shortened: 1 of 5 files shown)


piggz:
Looking good...
What happens if I select a random zip file?
What happens if i select a zip file containing manifest.json, but not the structure we expect?

JF002:
> Looking good... What happens if I select a random zip file? What happens if i select a zip file containing manifest.json, but not the structure we expect?

That's... something I'll need to check! Thanks for reminding me to also test error cases!

JF002:
So, I've just pushed a new commit [1db2d10](https://github.com/piggz/harbour-amazfish/pull/237/commits/1db2d103fd9beadf1b15d021dd4af87214d868ea) that checks for the validity of the archive and json file before starting the upload! The transfer will start only if all information are available in the archive and in the manifest file.

PR #238, opened 2022-09-05
Fix error management in DfuService (for PineTime/InfiniTime)

JF002:
If an error occurs during the OTA for the PineTime running InfiniTime, the variable `m_operationRunning` is still set to 1, which prevents the user from restarting the transfer.
This PR fixes that by adding a new signal `DfuOperation::transferError`. When emitted, the slot in `DfuService::onTransferError` correctly sets the variable to 0.

I also took the opportunity to store `m_updateFirmware` in a unique_ptr to ensure that it's correctly destroyed in every cases.


piggz:
Thanks!

What is ttt doing, and could (should?) it be moved to a member variable?

JF002:
ttt is probably a debug variable I forgot to remove. I'll check that (and fix conflicts) asap!

JF002:
I've just removed the debug code (the variable `ttt`) and rebased everything on master to fix the conflicts ;-)

PR #259, opened 2023-03-15
Refactor notification application icon

Thaodan:
The change should make it easier to add aliases for common apps that
either don't have icons or are related to those that have predefined
icons.


piggz:
thanks!

PR #270, opened 2023-09-05
Enable translations

### LINKED ISSUE #269: Does translations work?
I am wondering if translations has ever worked. I don't see usage of QTranslator class nor compilation of translation using lrelease in build process.

jmlich:
fixes #269 

This is blind implementation, I am not sure about behaviour in SailfishOS or other environments.

I have verified the behaviour at least in ubports branch
screenshot20230905_201006169


jmlich:
I have added also configuration file for integration with transifex. The integration must be done by github repository maintainer. After that the translations must be updated only as follows:
```
cd ./ui; lupdate -recursive . -ts ./translations/harbour-amazfish-ui.ts
```
for update of source file.
The translated files must be updated via transifex. You must upload .ts file into transifex when someone makes pull request to update language. Otherwise the transifex will not recognise that some change was made and it will replace changes with own copy in next synchronization.
The transifex integration can be configured to create pull requests or commits. Personally, I am using it in commits mode.

The transifex integration it self can be configured as follows:
- create an account + create an app
- settings
- integrations
- github
- allow access on github
- select repository
- configuration in transifex.yaml is used automatically

piggz:
This looks interesting... afaik translations were working fine in Sailfish ... atleast my translators are quick to provide updates, and I assume they work for them!

Ill probably accept this, but i need to think if transiflex will be best for my other translators, plus ive never used it so will need to read up.  thanks

eson57:
Translations is working perfectly fine, but problem is as often these days, provided rpm's comes with an empty translations directory, on chum and Open Repos both.  ;)

jmlich:
Now I see, the QTranslator is created with SailfishApp::application at
https://github.com/piggz/harbour-amazfish/blob/master/ui/src/harbour-amazfish-ui.cpp#L46

which makes QTranslator
https://github.com/sailfishos/libsailfishapp/blob/6336261392cbd55bdca41ee1c3f7ba424ce6a01f/src/sailfishapp_priv.cpp#L123

so the issue is specific only to non SailfishOS devices.

piggz:
One small thing
    DEFINES += TRANSLATION_FOLDER=\\\"$${PREFIX}/share/$${TARGET}/translations\\\"

is only set for silica ... what should this be for building  a kirigami version on a regular desktop build?

jmlich:
I am not sure what is standard for other platforms. It could be probably same as for sailfish

PR #271, opened 2023-09-22
Read battery level from PineTime

jmlich:
I was able to read battery information from pinetime.

See attached screenshot 

screenshot20230922_130234361


piggz:
lgtm, thanks :)

PR #272, opened 2023-09-29
Transliteration of notifications

jmlich:
It seems pinetime have troubles to draw non-ascii letters. Here, I am proposing solution for that. I still have some questions, but this proof of concept works 

On the attached photos I have received message "Čeština je řeč králů" (Czech language is language of kings) on telegram from Pavel Kerék. Not sure about other languages, but for me it works.

image20230929_093924783
image20230929_093930271

There are also other questions:
- is icu the right library? It seems to be dependency of Qt, hence I expect its availability on all platforms.
- where should be transliteration added? I have added it just for PineTime device into sendAlert() function, but same trouble will be with phone calls. It can be done also somewhere in libwatchface after getting the notification, or somewhere else yet.
- How is this solved with for other devices? Does they provide those characters?
- Is transliteration on smartphone the right way? It seems gadgetbridge have transliteration as configurable option. It seems to be recommended at https://github.com/InfiniTimeOrg/InfiniTime/issues/1490


piggz:
1) I think the ICU library is the correct solution for this.
2) Ill have to think about where to implement this.....I know the amazit devices support a range of languages, so maybe less of an issue there

piggz:
Ok, lets merge this as-is for now, and if we have other use cases, can expand its use later. thanks

PR #279, opened 2023-10-10
make transliteration configurable

### LINKED ISSUE #13: Umlauts are displayed as [?] (UTF)
I noticed Umlauts (and maybe other special characters) are displayed as ?.

jmlich:
fixes #13

screenshot20231010_125208306


jmlich:
It would be great to add reasonable defaults. For example enabled for PineTime.

jmlich:
I have reused function sendAlert in deviceInterface. I am not sure about intention while using sometimes m_device->sendAlert(...) and in other cases just sendAlert(...). It can be related to check of the same/repeated notification.

PR #286, opened 2023-10-17
There was missing first letter in notification

jmlich:
I am not sure if this works properly also on other smartwatches, but on PineTime this definitely helps. 

It seems generic alert message is used for phone call, but it is composed in different way. 
The alert is
category (one byte ) + alert_type=1 ( one byte ) + icon(one byte)  + sender (32 bytes) + separator (one byte) +subject (128bytes) + separator=\n (two bytes) + message (128bytes)

PineTime doesn't counts lenght sender+subject+message, together and clips it to 100 bytes

The incommingCall() composes message like this
03 01 + caller_id
I believe that header should be 3 bytes (I have added 0x01 for icon)
03 01 01 + caller_id

And finally  proof with picture:

image20231017_181829517
image20231017_181836912

The notification was send from debug dialog.

### DIFF
--- a/daemon/src/devices/huamidevice.cpp
+++ b/daemon/src/devices/huamidevice.cpp
@@ -137,7 +137,7 @@ void HuamiDevice::incomingCall(const QString &caller)
     qDebug() << Q_FUNC_INFO << caller;
     AlertNotificationService *alert = qobject_cast<AlertNotificationService*>(service(AlertNotificationService::UUID_SERVICE_ALERT_NOTIFICATION));
     if (alert) {
-        alert->incomingCall(caller);
+        alert->incomingCall(QByteArray::fromHex("0301"), caller);
     }
 }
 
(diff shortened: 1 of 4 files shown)


piggz:
Ive tested the original code with an Amazfit GTS, and it works as expected ... need to find the ANS spec!

jmlich:
Related documentation of InfiniTime can be found here
https://github.com/InfiniTimeOrg/InfiniTime/blob/main/doc/ble.md#notifications

piggz:
Yeah, i found that ... JF reminded me that the protocol hasnt changed since it was first committed ... and the person that wrote it was me! :laughing: 

Technically its -wrong- ... the spec is 2 bytes ... however ... its already implemented, so, im torn!

jmlich:
I hope, that provided patch makes it compatible with Amazfit

PR #289, opened 2023-10-20
Immediate Alert Service

jmlich:
I am playing with infinitime code and I need the phone part as well. I am making an pull request because it looks now quite good.

Sending immediate alerts to PineTime works quite good.  The Immediate Alert Service was previously used in 27b2b5ecaa1b25d0c9608389f26cda7ebc719be2 . By the way, this isn't working well for PineTime, because shows another notification instead of suppress voice call notification. I am not sure how to solve that.


jmlich:
I have moved unfinished parts into separate pull request. 
I am not sure, but it seems that some parts was already merged with some other changes.
The aim is:
- to bring immediate alert service with proper interface to be able reuse it in combination with other devices
- to bring testing button into "debug page".
- the interface should be used in onRingingChanged.

Issue #274, opened 2023-10-06
number of samples in mi_band_activity and heart rate activity and steps architecture


jmlich:
I am checking how can I implement steps and heart rate sampling for PineTime.

I can see that table have following structure
```
mi_band_activity(id, timestamp, timestamp_db, device_id,user_id, raw_intensity, steps, raw_kind, heartrate)
```

I can insert either sparse data like this:
```
for steps: (id, timestamp, timestamp_db, device_id, user_id, null, steps, null, null)
for heartrate (id, timestamp, timestamp_db, device_id, user_id, null, null, null, heartrate)
```
or I can create separate tables steps_log(id, timestamp, timestamp_db, steps) and heart rate(id, timestamp, timestamp_db, heart rate) and store data independently. This would expect either have every query twice in datasource.cpp or to migrate old data to new structures and adjust code accordingly.

Other option would be to collect both steps and heartrate in daemon and then push them into database after some period of time. This can reduce also the number of stored data or somehow normalize the output.

I am also wondering what amount of data is written every day. How this will affect lifespan of NAND? Can you please run following query and give me some hint. I am thinking of preventing storing samples more often than once per minute.
```
adb shell
sqlite3 /home/phablet/.local/share/harbour-amazfish/harbour-amazfish/amazfish.kexi
SELECT date(timestamp_dt), count(*) FROM mi_band_activity GROUP BY date(timestamp_dt) ;
```
It would be helpful to see if the data in mi_band_activity are actually sparse or dense
```
SELECT * FROM mi_band_activity
```

I am already playing with the implementation, but I would like to create it compatible with other devices.
![screenshot20231006_155834908](https://github.com/piggz/harbour-amazfish/assets/6171637/444ac048-c4d8-44b4-96c1-d8a51c20afab)

piggz:
sqlite> SELECT date(timestamp_dt), count(*) FROM mi_band_activity GROUP BY date(timestamp_dt) ;
2023-05-29|1038
2023-05-30|1440
2023-05-31|1440
2023-06-01|2358
2023-06-02|1166
2023-06-03|1440
2023-06-04|1440
2023-06-05|1440
2023-06-06|1440
2023-06-07|1440
2023-06-08|1440
2023-06-09|1440
2023-06-10|1440
2023-06-11|1360
2023-07-15|80
2023-07-16|1859
2023-07-17|1440
2023-07-18|1440
2023-07-19|1440
2023-07-20|1440
2023-07-21|1440
[...]

PR #303, opened 2023-10-28
Sampling of PineTime realtime activity and store it into mi_band_activity table

jmlich:
Until Infinitime implements physical activity monitor service (see https://github.com/InfiniTimeOrg/InfiniTime/issues/749) or similar mechanism to store and transfer sampled activity we can use only real time data from heart rate monitor and from motion service .

Closes https://github.com/piggz/harbour-amazfish/issues/274

### DIFF
--- a/daemon/daemon.pro
+++ b/daemon/daemon.pro
@@ -130,6 +130,7 @@ SOURCES += \
     src/typeconversion.cpp \
     src/bipbatteryinfo.cpp \
     src/devicefactory.cpp \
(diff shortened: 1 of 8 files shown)


jmlich:
The implementation is more less ready, I would just like to test it for few days.

jmlich:
The charts looks like this. They are sparse some time depending on the time when they are connected to device and sending data.

screenshot20231127_135842071
screenshot20231127_140810171
screenshot20231127_141552448

PR #306, opened 2023-10-29
Enable realtime HRM from Device Settings Page

jmlich:
Right now I am not sure how to do this properly.

With https://github.com/piggz/harbour-amazfish/pull/303 the sampled data are stored the realtime data are stored into database, but the notification of heart rate monitor must be enabled explicitly https://github.com/piggz/harbour-amazfish/blob/dc0aa13db6dbb8d5ffa7a89c16f847032e431fa9/daemon/src/deviceinterface.cpp#L741

There is enableRealtimeHRMeasurement and enableManualHRMeasurement 
https://github.com/piggz/harbour-amazfish/blob/dc0aa13db6dbb8d5ffa7a89c16f847032e431fa9/daemon/src/services/hrmservice.cpp#L37

both methods subscribes for messages. They are sending also some sending messages upon to infinitime. 

It can be probably some option in settings, which automatically subscribes to HRM upon connecting to device. In combination with lousy connection stability it doesn't measure very few data without that.


piggz:
Enable realtimeMeasurement tells the watch to enable its on-board periodic measurement (must be supported) ... enableManualMeasurement tells the watch to "take a measurement now please"

For the pinetime, i think the only current options is periodic manual measurement, as there is no feature for having the watch log measurements automatically.

jmlich:
Yes, the issue is, that without enableNotification() it doesn't receive any value.
On the First Page, the btnHR (with icon: styler.iconRefresh) must be pressed to turn it on. 

If I understand correctly the right method is `enableRealtimeHRMeasurement` which is currently used in combination with Laufhelden app (or other app connected via dbus daemon). I guess, I want to introduce some TextSwitchPL on DeviceSettings page and enable this automatically according to it.

jmlich:
This should allow realtime HRM measurement (i.e. notifications). 
As far as PineTime doesn't provide interface to download hrm logs, this should allow at least sampling of current values. 

Dark side is that it is necessary to allow it in Settings. The user can be confused that data aren't actually there. On the other hand user can be surprised that it drains device battery (haven't measured that actually)  and doesn't bring much changes (mostly for huami device).

Although I see the pull request as an improvement.

ui/qml/pages/Settings-device.qml
```diff
@@ -121,6 +130,7 @@ PagePL {
                 DaemonInterfaceInstance.applyDeviceSetting(Amazfish.SETTING_DEVICE_TIME);
                 DaemonInterfaceInstance.applyDeviceSetting(Amazfish.SETTING_DEVICE_UNIT);
                 DaemonInterfaceInstance.applyDeviceSetting(Amazfish.SETTING_DISCONNECT_NOTIFICATION);
+                DaemonInterfaceInstance.applyDeviceSetting(SETTING_DEVICE_REALTIME_HRM_MEASUREMENT);
```
piggz, line 133: Syntax here is different

jmlich: fixed

PR #309, opened 2023-11-02
Battery log stored in generic table

jmlich:
The main difference against https://github.com/piggz/harbour-amazfish/pull/273 is storing both key and value.

There are still some questions:
- what other values should be stored in this way? Personally I prefer implementation in 273 until other key isn't present.
- there is an loading of last value feature https://github.com/piggz/harbour-amazfish/pull/273/commits/9ea2f53a894ac6e7454e332bf10f2a92d39b7c0d to avoid duplicating value after restart (which is mainly case of debugging which can cause many extra rows in table and leads to ugly graph). Should be this feature extended to read all last value for all keys  in info_log?


jmlich:
btw I don't like redundancy of Info enum

https://github.com/piggz/harbour-amazfish/blob/cf13e026b15806b2b1ad55375e831f9bdee41ae5/lib/src/amazfish.h#L46-L60
https://github.com/piggz/harbour-amazfish/blob/cf13e026b15806b2b1ad55375e831f9bdee41ae5/daemon/src/devices/abstractdevice.h#L27-L40

piggz:
Yeah, i think there was a reason for that, but it escapes me now!

jmlich:
The issue with redundant enum isn't something I wanted to address in this pull request. It was mentioned just out of curiosity.

PR #318, opened 2023-12-08
SailfishOS build using Github Actions

### LINKED ISSUE #313: Continuous Integration
I was wondering how to check if the package builds for other platforms. It seems to be reasonable to use some CI.
There are more options and I am not sure which one is prefered:

- Github actions
Here is some discussion:
https://forum.sailfishos.org/t/build-sailfish-os-application-using-github-actions-or-gitlab-ci/2147/18
Here is some example:
https://github.com/CODeRUS/screencast/blob/master/.github/workflows/build.yml

It seems the clickable (ubuntu touch) container is larger then github actions provides for free.

- Chum repository
I am not sure if this is suitable also for build of all branches.
I have found this one
https://build.sailfishos.org/project/show/sailfishos:chum:testing

- Something else
There are many other options to take.

jmlich:
- verifies build of sailfishos package
- creates RPM packages for testing

Fixes #313 

I am not able to verify if are packages installable because I don't have SailfishOS device

### DIFF
--- a/.github/workflows/sailfishos.yml
+++ b/.github/workflows/sailfishos.yml
@@ -1,0 +1,38 @@
+name: SailfishOS Build
+
+on: [push, pull_request]
+
+env:
+  OS_VERSION: 4.5.0.16
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    name: Build App
+    strategy:
+      matrix:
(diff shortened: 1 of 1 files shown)


piggz:
I tested the aarch64 version on device and it works

PR #319, opened 2023-12-12
Infinitime Simple Weather Service

### LINKED ISSUE #308: Weather Service rework
We would like to rework weather service to be simpler. It should work according following description:

> simple BLE API : it currently just take a 6 bytes buffer:
>  [0] = PacketType (only the value "0" is implemented, but this will allow for extensions in the future)
>  [1][2] = Temperature (16bits)
>  [3] = Precipitation type
>  [4] = Precipitation amount
>  [5] = Cloud amount
> 
> and... that's it! Amazfish just has to send the current values whenever they are updated.

jmlich:
This is was created in cooperation with @JF002 in scope of https://github.com/piggz/harbour-amazfish/issues/308. (Closes #308)
Documentation of the service and reasons are described in InfiniTime pull request.
https://github.com/InfiniTimeOrg/InfiniTime/pull/1924

I have update the osm weather in order to provide icon and temperature in qreal, because it converted from kelvins which adds unwanted error to the value (for example 274.9 kelvins ~ 274 - 273 = 1 deg, but it should be 274.9 - 273.15 = 1.75 ~ 2 deg). The weather icon is displayed in cityselection dialog.

### DIFF
--- a/.github/workflows/sailfishos.yml
+++ b/.github/workflows/sailfishos.yml
@@ -1,0 +1,38 @@
+name: SailfishOS Build
+
+on: [push, pull_request]
+
+env:
+  OS_VERSION: 4.5.0.16
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    name: Build App
+    strategy:
+      matrix:
+        arch: ['armv7hl', 'aarch64', 'i486']
+
+    steps:
+    - uses: actions/checkout@v3
+      with:
+        submodules: recursive
+    - name: Prepare
+      run: docker pull coderus/sailfishos-platform-sdk:$OS_VERSION && mkdir output
+
+    - name: Build ${{ matrix.arch }}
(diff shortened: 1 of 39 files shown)


JF002:
We merged the new Simple Weather Service in InfiniTime. I did most of my tests using this PR in Amazfish, so I guess it's pretty safe to merge it ;-)
Thanks again a lot to @jmlich for their help on this topic!

jmlich:
There is a few things to consider
- I have modified the openweather api to get temperature in float, not int. The floating point is better to keep precision while converting to Fahrenheit scale. I was looking into code and it should work for other devices, but I will be happier if you can check if it really works.
- Both weather services are still there. New InfiniTime just doesn't advertise the other GATT service, so it is created just right instance of Weather Class. For Old InfiniTime is it other way around. 
- I did some changes in weather selection page UI. It shows the icon now. Does it look fine on SailfishOS?

jmlich:
I should probably rename the service to pinetimesimpleweatherservice to highlight that this is pinetime specific.

piggz:
Good idea

Issue #284, opened 2023-10-15
Add support for AsteroidOS watches


sviscapi:
Dear Adam,

Could you please add support for AsteroidOS watches to Amazfish ?

https://asteroidos.org

I tried https://github.com/AsteroidOS/starship but couldn't find a binary compatible with Sailfish OS 4.5 on aarch64.

Best regards,

Samuel

jmlich:
From the documentation https://wiki.asteroidos.org/index.php/BLE_profiles it seems they are using different bluetooth services then was implemented in amazfish. Basically only battery service is the same. In theory, the libasteroid could be adjusted (and it doesn't look that difficult). In real world, some device is needed to debug it. I am not able to help with that unless you want to donate some device.

RelevanterUsername:
as little contribution from my side I would like to point out that asteroidos support has been implemented in gadgetbridge and that code from there possibly could be used: [](https://codeberg.org/Freeyourgadget/Gadgetbridge/pulls/3013)

robthebold:
> From the documentation https://wiki.asteroidos.org/index.php/BLE_profiles it seems they are using different bluetooth services then was implemented in amazfish. Basically only battery service is the same. In theory, the libasteroid could be adjusted (and it doesn't look that difficult). In real world, some device is needed to debug it. I am not able to help with that unless you want to donate some device.

I've got a spare device, if you're somewhere I can get it shipped without spending more than the thing is worth. (The watch lives in USA right now.)  Lemme know if interested.

jmlich:
I am in Czech Republic (EU). Please contact me at jmlich83@gmail.com

jmlich:
I have first step of implementation here https://github.com/jmlich/harbour-amazfish/tree/asteroidos However, I am not able to debug it without device. There is only Battery Service right now, so don't expect too much. I even don't know if this works properly.

RelevanterUsername:
I could test this, but I don't know how to install it. IS there a compiled rpm package or only the source available?

jmlich:
I am making builds just for Ubuntu Touch. Currently, I don't have SailfishOS SDK installed. I guess some CI may help in that area https://github.com/piggz/harbour-amazfish/issues/313

RelevanterUsername:
I installed armv7 on my device, but cannot connect to watch (sturgeon with asteroidos). What is the correct order of doing the pairing? Or isn't this supposed to be working yet?

jmlich:
[...]

PR #320, opened 2023-12-24
AsteroidOS support

jmlich:
Support for AsteroidOS. 

- [x] the pairing page introduced aliases. The aim was to reduce number of devices in the list. There is about 20 code names for asteroid os devices
- [x] battery service is generic
- [x] weather service
- [x] time service 
- [x] notification service
- [x] media service

Closes https://github.com/piggz/harbour-amazfish/issues/284


jmlich:
- I am still not sure what to do with authentication agent, but it somehow works when is device paired in os in the same time. 
- The Screenshot feature requires QBLEDescriptor. Additionally, I have also some minor bug there. It will be better to add it in separate pull request.
- The volume control in Media Service doesn't work perfectly. Some change in pulseaudio/libwatchfish is needed. This can be done in future.

piggz:
Well, as I cant test this, and it looks reasonable, lets merge ... id like the large commented sections removed later :)

PR #329, opened 2024-01-13
Update Chum Metadata to new format

nephros:
while on the topic of updating things for chum, lets use the newer metadata format.


jmlich:
It looks good to me. It matches fields described at https://github.com/sailfishos-chum/main/blob/main/Metadata.md

Donation field can be probably added into Links section.
```
  Donation: https://www.paypal.me/piggz
```

PR #336, opened 2024-01-21
AsteroidOS screenshot

jmlich:
there are few issues so far:
- [x] file contain few extra bytes and screenshot looks corrupted
- [x] files should be stored probably into ~/Pictures
- progress bar is not show in ui (it is not necessary, but poor UX)


jmlich:
I have spent a lot of time with QDataStream which adds own header.

daemon/src/devices/asteroidosdevice.cpp
```diff
@@ -300,22 +300,19 @@ void AsteroidOSDevice::screenshotReceived(QByteArray data) {
     }
 
     QString fullpath = picturelocation.absolutePath() + "/" + filename;
-    QFile *screenshotFile = new QFile(fullpath);
+    QFile screenshotFile(fullpath);
 
-    qDebug() << fullpath;
+    qDebug() << fullpath << data.size();
 
     QDataStream *dataStream = nullptr;
```
piggz, line 307: Has this been left over?

jmlich: I think file name and size is useful debug message.

piggz:
Is this ready, or are you wanting to add the progress bar part?

jmlich:
I don't want to work on progress bar. It is nice to have feature, but it requires many changes in daemon.

piggz:
Should we have an additional FEATURE flag for this FEATURE_SCREENSHOT, so that the UI can update with whether or not it is available?

jmlich:
flag has been added

PR #337, opened 2024-01-25
[Cleanup/Debug Logs] Use  Q_FUNC_INFO in many more places.

nephros:
Make debug info and include guard match source file/function names

### DIFF
--- a/daemon/src/activitysummary.cpp
+++ b/daemon/src/activitysummary.cpp
@@ -172,18 +172,18 @@ bool ActivitySummary::saveToDatabase(KDbConnection *conn)
     values << baseAltitude();
     values << m_gpxTrack;
 
-    qDebug() << "Saving Sport:" << name()  << startTime() << startTime().toString(Qt::ISODate);
+    qDebug() << Q_FUNC_INFO << "Saving Sport:" << name()  << startTime() << startTime().toString(Qt::ISODate);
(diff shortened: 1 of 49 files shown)


piggz:
Thanks! ... could you use Q_FUNC_INFO instead of strings ... i generally do this nowadays :)

jmlich:
In general, I like that change and I like having function name in the log. However, there is a generic way to do that.
The function myMessageOutput  can be adjusted to provide all the details.

https://github.com/piggz/harbour-amazfish/blob/d5e3fe6b67b21ebe0710d8abc39dca83f04b4d06/daemon/src/harbour-amazfish-daemon.cpp#L25-L30


I am using that in few projects, for example:
https://github.com/jmlich/geotagging/blob/a332aa865e0e4eae655fa8f998bc63c8ba4a089a/main.cpp#L9-L82

It have some pitfalls, e.g. QT_MESSAGELOGCONTEXT define and some compiler flag must be used, otherwise filename and line is removed.

jmlich:
Its not request, it maybe suggestion or opening of discussion. As I said previously, I am more less fine with keeping it same as you made it.

piggz:
Ill accept this for now ... ive used the myMessageHandler approach in the past, and maybe something to do as a next step?

PR #344, opened 2024-02-25
Banglejs : Improvement over notifications

Tsht:
1. Incoming calls

From specifications of notifications of gadgetbridge (https://www.espruino.com/Gadgetbridge), cmd can be 
t:"call", cmd:"accept/incoming/outgoing/reject/start/end", name: "name", number: "+491234" - call

example shows :
GB({"t":"call","cmd":"incoming","name":"name","number":"+441234123123"})

Tried this way and incoming calls show on my bangle.js 2
I didn't check however how to accept or refuse calls...

2. Notifications (sendAlert)

Put an unique identifier.
Also, filter out messages which have an empty body and subject. Not convinced 100%, but something similar is done for instance for MiBandService::sendAlert
For me, it concerns mails which sends two notifications, the second with "courriel" (so in French) in the sender field, with empty subject and message.

I suppose it would be better to filter out those based on the event, but I do not master this part of code.

### DIFF
--- a/daemon/src/devices/banglejsdevice.cpp
+++ b/daemon/src/devices/banglejsdevice.cpp
@@ -51,9 +51,30 @@ void BangleJSDevice::sendAlert(const QString &sender, const QString &subject, co
         return;
     }
 
+    //For mails, there is a double notification which is empty except for sender
+    if(message.isEmpty() && subject.isEmpty())
+    {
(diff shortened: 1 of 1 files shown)


piggz:
thank! hopefully you can make more bangle.js improvements

PR #346, opened 2024-03-01
Pop after timer refresh daemon

### LINKED ISSUE #339: Alarms not staying set
Can enable and save an alarm.  It shows as enabled but the watch never buzzes to wake you up in the morning.If you check the alarm setting later on during the night it doesn't show as being set. This used to work and only seems to have stopped since the January update.

jmlich:
Fixes #339 

It seems app.pages.pop() calls destructor of the Page including Timer before it is triggered


piggz:
thanks, was going to to this myself but you beat me to it!

PR #357, opened 2024-04-05
AmazfitNeo notification is using AlertNotificationService

jmlich:
it seems that max length of message is limited.
The specification says 18 characters. The screen shows only "+1" or "+2" anyway.

Code worked for me, but it would be great if @kirbylife can look at it as well.

### DIFF
--- a/daemon/src/devices/huamidevice.h
+++ b/daemon/src/devices/huamidevice.h
@@ -27,7 +27,7 @@ public:
 
     void applyDeviceSetting(Settings s) override;
 
-    void sendAlert(const QString &sender, const QString &subject, const QString &message) override;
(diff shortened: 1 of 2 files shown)


kirbylife:
Yes, it seems to work as it would on Android.
The only issue I see is that there is no way to enable/disable that feature from Amazfish, I had to go to GadgetBridge to enable notifications and then go back to Amazfish to test.
I don't know if I didn't find the way to do it or if it is expected.

jmlich:
I don't think that there is such configuration in Amazfish. However, it can added.

kirbylife:
I liked the way you implemented it.
Adding the option to enable/disable notifications could go in another PR.

PR #380, opened 2024-05-17
[kirigami] DatePicker dialog doesn't pop() when accepted

jmlich:
In c48f79b96e6e81e2e01f60a0c36ec7706d6887b6 there was a fix of Dialog behaviour for Device Unpair operation. This broke DatePicker Dialog (User Settings -> Birthday) .

In this fix is used acceptDestination variable show the Unpair works with previous implementation and DatePicker works as well.

Same issue applies to TimePicker Dialog as well.

The change may affect also unpair page

### DIFF
--- a/ui/qml/components/platform.kirigami/DialogPL.qml
+++ b/ui/qml/components/platform.kirigami/DialogPL.qml
@@ -38,9 +38,9 @@ PagePL {
             if (acceptDestinationPop)
                 app.pages.pop(acceptDestination);
             else
-                app.push(acceptDestination);
-        } //else
-            //app.pages.pop();
+                app.pages.push(acceptDestination);
+        } else
+            app.pages.pop();
     }
 
     function accept() {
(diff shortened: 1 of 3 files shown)


piggz:
is this still wip?

jmlich:
Yes, there are at least two cases with use of that mechanism which are bound together. I not sure exactly, therefore mentioning following three (maybe I have missed something):

- SettingsProfile -> Birthday (DatePickerDialogPL) -> Save()
- FirstPage -> PairSelectDeviceType -> PairPage -> FirstPage
- FirstPage -> UnpairDeviceDialog -> PairSelectDeviceType -> PairPage -> FirstPage

I have seen some suspicious behavior with UnPairDialog, where it should replace dialog or pass it with attachedpage mechanism, but it was somehow unreliable. I have seen overlapped FirstPage in background in SailfishOS.

I have spent some time with that already without reasonable result.

jmlich:
It looks like everything works now on all platforms. I am wondering if there were another scenario which I didn't considered or forgot.

PR #409, opened 2024-10-10
add total sleep time and reformat sleep times

Danfro:
I find it confusing to see the "light" sleep time in very large letters and the small deep sleep time and those not matching the total time on the graph. I think more interesting is the total time to answer the question "how long did I sleep".

This PR will add total sleep label, name light sleep text with "light" and format all times with the locale short format.

Before:

grafik

After:

grafik

### DIFF
--- a/ui/qml/pages/SleepPage.qml
+++ b/ui/qml/pages/SleepPage.qml
@@ -19,14 +19,23 @@ PagePL {
         width: parent.width
         anchors.top: parent.top
         anchors.margins: styler.themePaddingMedium
-        spacing: styler.themePaddingLarge
+        spacing: styler.themePaddingMedium
 
         LabelPL {
             id: lblSleepLastnight
-            font.pixelSize: styler.themeFontSizeExtraLarge * 3
+            font.pixelSize: styler.themeFontSizeExtraLarge * 2
             anchors.horizontalCenter: parent.horizontalCenter
             width: parent.width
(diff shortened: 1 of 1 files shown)


Danfro:
I should add, I only use Ubuntu Touch, so I have no idea (and can't test) on Sailfish OS.

jmlich:
I don't like that `decimalToHourMin` returns '12:00 AM'. 

SailfishOS and Kirigami looks fine:

Screenshot-24-10-11-12-35-14
Screenshot at 2024-10-11 12-33-45

Danfro:
Thanks a lot for testing. And of course you are right, the am/pm need
to be omitted.

I guess a manual build string is the best option here. At least from a
first search I can't find a good solution for toLocaleTimeString except
of adding a "replace("am","").replace("pm","").

If we use a manually build string, we could have it either "12:45"
notation or "12 h 45 min" notation. Which would you prefer?

jmlich:
I suggest to use this:
```
function decimalToHourMin(decHours) {
    var totalMinutes = Math.floor(decHours * 60);
    var hours = Math.floor(totalMinutes / 60);
    var minutes = (totalMinutes % 60).toString().padStart(2, '0');
    
    return hours + ":" + minutes;
}
```

Danfro:
I am sorry, I didn't check the previous comments and forgot about that. Yes, something pretty much like that I had in mind. Thank you very much for that snippet. That makes this change even more convenient. 

I just did add that with two small changes:
1. rename decHours to decTime since I think that is more correct
2.  use math.round instead of math.floor for the totalMinutes

jmlich:
Looks good to me now.

PR #410, opened 2024-10-11
SailfishOS: enable cover page

jmlich:
It seems the SailfishOS cover page was disabled some time ago. It seems it have useful information to show and allows more user interaction.
Screenshot-24-10-11-11-27-40

The property was added in order to allow use it with other flavors. 
It was necessary to fix images path. The `styler.iconHeartrate` contain one more level so I have decided to use absolute path for now.


piggz:
lgtm, thanks!

PR #411, opened 2024-10-14
Update build instructions

Danfro:
This PR will update the repo links for the two sailfish plugins. It will also extend the note for UT build instructions to mention the libraries, since the command given will build both, the app and the libs.

Important for myself was, that the libs need to be build against the target architecture e.g. arm64. So this command needs to be run with a device attached. Or libs need to be build manually by specifying the correct architecture. Simply building will use the hosts arch which will fail. I hope my last rephrasing makes that clearer.


jmlich (approved):
lgtm

PR #413, opened 2024-10-14
rename download to upload since files are uploaded onto the device

Danfro:
I am not sure if I do miss something, but this menu and function seems to upload files to the watch, not load files from the watch. So I suggest renaming it to upload.

### DIFF
--- a/ui/qml/pages/BipFirmwarePage.qml
+++ b/ui/qml/pages/BipFirmwarePage.qml
@@ -4,7 +4,7 @@ import "../components/platform"
 
 PagePL {
     id: page
-    title: qsTr("Download File")
+    title: qsTr("Upload File")
 
     property string selectedFile: qsTr("None")
     property string fileVersion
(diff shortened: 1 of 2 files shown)


jmlich:
+1
I must admit, `download` label was very confusing for me, but I used to it ;-D

piggz:
How about "Install" instead?

Danfro:
I only started using the watch and the app recently, so I was not 100% sure if "uploading" is always "installing" or might be "uploading" something else too. I wouldn't mind "installing", I guess that works as well. @jmlich any preference here? Upload or Install?

jmlich:
PineTime allows to install firmware and to upload some resources (navigation icons)

Danfro:
One could consider "uploading of icons" as "installing icons".

And thinking about it, maybe "install file" is better, because it warns the uer, that they are maybe going to change something on the system that can't easily be reverted. If no one is against it, I can change the PR later to adapt.

--

jmlich:
Btw there is an page explaining how install resources with Amazfish screenshots https://github.com/InfiniTimeOrg/InfiniTime/blob/main/doc/gettingStarted/updating-software.md#updating-resources

PR #415, opened 2024-10-14
show locale number format for steps

Danfro:
This PR will use the same locale number format for steps as on [FirstPage](https://github.com/piggz/harbour-amazfish/blob/master/ui/qml/pages/FirstPage.qml#L186). 

It will also drop showing the steps of the last day when steps are zero. I personally find it confusing to see a number there if I know I didn't do any steps. On other pages also zero is shown if there is no reading (yet). So I suggest dropping that part and show zero steps if there are zero steps.

### DIFF
--- a/ui/qml/pages/StepsPage.qml
+++ b/ui/qml/pages/StepsPage.qml
@@ -25,7 +25,7 @@ PagePL {
             font.pixelSize: styler.themeFontSizeExtraLarge * 3
             anchors.horizontalCenter: parent.horizontalCenter
             width: parent.width
-            text: _InfoSteps > 0 ? _InfoSteps : graphStepSummary.lastValue
+            text: _InfoSteps.toLocaleString()
             horizontalAlignment: Text.AlignHCenter
         }
 
--- a/ui/qml/pages/StepsPage.qml
+++ b/ui/qml/pages/StepsPage.qml
@@ -25,7 +25,7 @@ PagePL {
             font.pixelSize: styler.themeFontSizeExtraLarge * 3
             anchors.horizontalCenter: parent.horizontalCenter
             width: parent.width
-            text: text: _InfoSteps > 0 ? _InfoSteps.toLocaleString() : graphStepSummary.lastValue
+            text: _InfoSteps > 0 ? _InfoSteps.toLocaleString() : graphStepSummary.lastValue
             horizontalAlignment: Text.AlignHCenter
         }
 
--- a/ui/qml/pages/StepsPage.qml
+++ b/ui/qml/pages/StepsPage.qml
@@ -25,7 +25,7 @@ PagePL {
             font.pixelSize: styler.themeFontSizeExtraLarge * 3
             anchors.horizontalCenter: parent.horizontalCenter
             width: parent.width
-            text: _InfoSteps > 0 ? _InfoSteps.toLocaleString() : graphStepSummary.lastValue
+            text: _InfoSteps > 0 ? _InfoSteps.toLocaleString() : graphStepSummary.lastValue.toLocaleString()
             horizontalAlignment: Text.AlignHCenter
         }
 
--- a/ui/qml/pages/StepsPage.qml
+++ b/ui/qml/pages/StepsPage.qml
@@ -25,7 +25,7 @@ PagePL {
             font.pixelSize: styler.themeFontSizeExtraLarge * 3
             anchors.horizontalCenter: parent.horizontalCenter
             width: parent.width
-            text: _InfoSteps.toLocaleString()
+            text: text: _InfoSteps > 0 ? _InfoSteps.toLocaleString() : graphStepSummary.lastValue
             horizontalAlignment: Text.AlignHCenter
         }
 


jmlich:
I guess `graphStepSummary.lastValue` takes last row from database. This is suitable when smartwatch are out of range or disconnected for some reason.

Danfro:
As far as I observed it, yes,  `graphStepSummary.lastValue` takes last days value from database. But I think this confusing to see old values where on other pages like sleep or heartrate always zero is given.

But maybe as a fail catch, we could do: `_InfoSteps > 0 ? _InfoSteps.toLocaleString() : 0`?

Also setting this value to current steps only and omitting the last days reading doesn't trigger any communication, does it? At least I can't see any obvious "on value changed" that triggers a data flow.

jmlich:
Maybe we should update other pages to use value from database as well.
I think this 
```
text: _InfoSteps.toLocaleString()
``` 
is better than this 
```
text: _InfoSteps > 0 ? _InfoSteps.toLocaleString() : 0
```
, because IMHO it will show the same value.

I am not sure if following syntax is possible, but I would prefer something like this.
```
text: (_InfoSteps > 0 ? _InfoSteps : graphStepSummary.lastValue).toLocaleString()
```

Danfro:
> Maybe we should update other pages to use value from database as well. 

Definitely not in my opinion. Every day the value for that day is given. So why would I want to display for the current day the value of the last day? In most cases zero is the correct value for that day, e.g. zero hours sleep, zero steps walked, ... .

So unless we have no value (or zero is wrong for any other reason I can't think of right now), then maybe better do what the graph does, display "no data".


I think this
> 
> ```
> text: _InfoSteps.toLocaleString()
> ```
> 
> is better than this
> 
> ```
> text: _InfoSteps > 0 ? _InfoSteps.toLocaleString() : 0
> ```
> 
> , because IMHO it will show the same value.

I didn't test if `_InfoSteps` can be `undefined` and we might need to cover that case. But if it is zero as default, then I do agree, the first solution is the better one.

We might want to do something like this maybe: `text: _InfoSteps  ? _InfoSteps.toLocaleString() : "no data"`?

> I am not sure if following syntax is possible, but I would prefer something like this.
> 
> ```
> text: (_InfoSteps > 0 ? _InfoSteps : graphStepSummary.lastValue).toLocaleString()
> ```

This would need to be  `text: (_InfoSteps > 0 ? _InfoSteps.toLocaleString() : graphStepSummary.lastValue).toLocaleString()`, but as said above, I don't like this at all because it shows a "wrong" value.

> > Also setting this value to current steps only and omitting the last days reading doesn't trigger any communication, does it? At least I can't see any obvious "on value changed" that triggers a data flow.
> 
> I am sorry for confusion, the #414 will trigger communication with device. I wanted to say, that current syntax takes information from database without need of that communication.

#414 triggers the init in FirstPage, not here in StepsPage. Without #414 the only init I found is the onCompleted in StepsPage. That means a user needs to leave FirstPage and enter StepsPage and return to get the current value in FirstPage. That seems not good to me, hence the MR. And there I used the `_connected` property to make sure to only trigger it if a device is connected. 

StepsPage triggers the init of `_InfoSteps` in it's onCompleted event anyway (without any condition), so I can't see this MR adding any dataflow that isn't already there or see any reason why to use a databasevalue.

jmlich:
> ... `text: _InfoSteps.toLocaleString()` ... 

This one is enough, because there is https://github.com/piggz/harbour-amazfish/blob/b3fe7cc268ef343786e513064f5dbe257c258d6f/ui/qml/harbour-amazfish.qml#L200
and default zero at https://github.com/piggz/harbour-amazfish/blob/b3fe7cc268ef343786e513064f5dbe257c258d6f/ui/qml/harbour-amazfish.qml#L36

> https://github.com/piggz/harbour-amazfish/pull/414 triggers the init in FirstPage, not here in StepsPage. Without https://github.com/piggz/harbour-amazfish/pull/414 the only init I found is the onCompleted in StepsPage. That means a user needs to leave FirstPage and enter StepsPage and return to get the current value in FirstPage. That seems not good to me, hence the MR. And there I used the _connected property to make sure to only trigger it if a device is connected.
> 
> StepsPage triggers the init of _InfoSteps in it's onCompleted event anyway (without any condition), so I can't see this MR adding any dataflow that isn't already there or see any reason why to use a databasevalue.

The _InfoSteps seems to be global variable used in both FirstPage and StepsPage.
Additionally, I am believe the _connected shouldn't be used here. I may be wrong, but if I understood correctly, the flow of information is something like this:
- AmazfishUI calls refreshInformation()
- Daemon does if (m_device) m_device->refreshInformation();, but the m_device is null when disconnected
 - m_device->refreshInformation() is device specific, but leads to InformationChanged
 - it leads to something like if (motionService) motionService->refreshSteps(), the BLE service is not created when not connected.
- Then Amazfish UI gets value in onInformationChanged: and stores into _InfoSteps variable

I believe it doesn't reconnect automatically.



>>    Maybe we should update other pages to use value from database as well.

> Definitely not in my opinion. Every day the value for that day is given. So why would I want to display for the current day the value of the last day? In most cases zero is the correct value for that day, e.g. zero hours sleep, zero steps walked, ... .

The issue probably is that the daemon doesn't provide interface to get last steps value, and streams value immediately when it comes from bluez. E.g bangle.js send steps counter information once per 10 minutes. It can take up 10 minutes to see fresh information in UI. Last information from database seems to be better then showing 0 or "no data".

jmlich:
If you want just to use locale number format for steps then it is that code:
```
 text: _InfoSteps > 0 ? _InfoSteps.toLocaleString() : graphStepSummary.lastValue.toLocaleString()
```
otherwise you should probably change the title to something like `don't use last value from database`

jmlich (commented):
copy paste trouble
/s/text: text:/text:/

Danfro:
@jmlich thanks for pointing that mistake out. Fixed. The apparently simplest things most often have mistakes in them... I need to be more careful.

jmlich (approved):

Danfro:
@jmlich due to our long conversation I forgot to add the locale format also to the database value. That works as well on my device. Just so you can reconsider your approval after my additional change.

jmlich (approved):

PR #418, opened 2024-10-24
add column with bpm tresholds and use round instead of floor

Danfro:
This PR will do two things in heart rate page:
1. it will use round instead of floor to calculate the category values to avoid larger differences when summing all up (sum should be 100%, but with floor can be easily 5% less, now at most 1% rounding difference)
2. it will add a third column with the actual threshold values which are used to calculate the categories, those values are important in my opinion, because a user might want to set a bpm warning on their watch (bip allows to do that) when exercising, but for that they need to now the value

grafik

### DIFF
--- a/ui/qml/pages/HeartratePage.qml
+++ b/ui/qml/pages/HeartratePage.qml
@@ -85,44 +85,64 @@ PagePL {
 
         //Type summary
         Grid {
-            columns: 2
+            columns: 3
             spacing: styler.themePaddingMedium
-            width: parent.width
-            LabelPL { text: qsTr("Relaxed") }
-            Item { width: parent.width * 0.5; height: 50
+            width: parent.width - (styler.themePaddingMedium * 2)
+            LabelPL {text: qsTr("Relaxed")}
+            Item { 
+                width: parent.width * 0.5
+                height: 50
                 Rectangle { color: "grey"; width: parent.width * (relaxed  / total) ; height: parent.height }
-                LabelPL { text: Math.floor((relaxed / total) * 100) + "%"; anchors.centerIn: parent}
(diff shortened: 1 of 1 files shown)


ui/qml/pages/HeartratePage.qml
```diff
@@ -85,45 +85,65 @@ PagePL {
...
+            Item { 
+                width: parent.width * 0.5
+                height: 50
                 Rectangle { color: "grey"; width: parent.width * (relaxed  / total) ; height: parent.height }
-                LabelPL { text: Math.floor((relaxed / total) * 100) + "%"; anchors.centerIn: parent}
+                LabelPL { text: Math.round((relaxed / total) * 100) + "%"; anchors.centerIn: parent}
             }
-            LabelPL { text: qsTr("Light") }
-            Item { width: parent.width * 0.5; height: 50
+            LabelPL {text: qsTr("<= %1 BPM".arg(Math.round(maxHR()*0.5)))}
```
jmlich, line 98: I would prefer use of ≤ symbol.
```
-            LabelPL {text: qsTr("<= %1 BPM".arg(Math.round(maxHR()*0.5)))}
+            LabelPL {text: qsTr("≤ %1 BPM".arg(Math.round(maxHR()*0.5)))}
```

Danfro: Thanks for that suggestion. Yes, that looks much better. :+1: 

grafik

jmlich (approved):

piggz:
thankyou, great to have all these improvements from you!

Danfro:
You are welcome. Just picking some low hanging fruit. And its always easier to fix small things than coding a full stack app in the first place.

PR #419, opened 2024-10-24
replace Haskell & Fox maxHR formula with more precise Wingate formula

Danfro:
The current formula to calculate the maxHR is the oldest and least accurate one. Its also based on a very small number of observations. Although it is mentioned on pretty much all pages on that topic, and its not utterly wrong, most websites agree that this is not the best solution.

The current implementation also has one big flaw: It doesn't cover the case when no date of birth = age is provided in the user data. Then the maxHR is used without any subtractions (age = 0). This leads to a very high maxHR and in the worst case someone may be over straining themself. That's why I think something should be changed.

I did some reading on this, and [Wikipedia](https://en.wikipedia.org/wiki/Heart_rate#Maximum_heart_rate) actually gives a good overview. Interesting, on German webpages, the Sally Edwards formula is among the most popular ones, which isn't even mentioned in the English wiki page. [Wikipedia German page](https://de.wikipedia.org/wiki/Maximalpuls#Weitere_Formeln)

Although I do quite like the idea of not only adjusting the maxHR by gender and age, but also by weight (Sally Edwards). But from what I read, this is not providing a significant improvement. So I did not follow that line.

In the end I was undecided between the Oakland University formula and the Wingate one.

Wingate:
+ newest research
+ based on a huge number of data compared to all the others
+ adjusts for gender and age
- linear only

Oakland University -  nonlinear formula:
+ smallest confidence interval
+ nonlinear age adjustment
+ longitudinal observations
- small number of observations
- a few years older than wingate
- no gender adjustment

I decided for the Wingate formula, mainly because of the much larger dataset its based on and because it allows gender based adjustments. I also added a default age of 50 if not set. This seems to be a) average enough and b) young people have generally a better health, so we rather want to provide a value a bit lower than too high.

I did play around with the numbers in a simple spreadsheet. I can't upload that here, that's why I drop the screenshot to provide at least a rough idea.

grafik

### DIFF
--- a/ui/qml/pages/BipFirmwarePage.qml
+++ b/ui/qml/pages/BipFirmwarePage.qml
@@ -4,7 +4,7 @@ import "../components/platform"
 
 PagePL {
     id: page
-    title: qsTr("Download File")
+    title: qsTr("Upload File")
 
     property string selectedFile: qsTr("None")
     property string fileVersion
@@ -19,7 +19,7 @@ PagePL {
 
         LabelPL {
             width: parent.width
-            text: qsTr("Select a file to download.");
+            text: qsTr("Select a file to upload.");
         }
 
         ValueButtonPL {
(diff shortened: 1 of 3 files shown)


jmlich (approved):

piggz:
Thankyou!

PR #435, opened 2024-12-13
Remove unused btdiscoverymodel.{cpp,h}

jmlich:
Those files are not in daemon.pro. It seems to be leftover of some previous refactoring.


jmlich:
Similar `./src/services/infinitimeresourceservice.*` seems to be incomplete, not included into build.

PR #440, opened 2024-12-24
onUnitNew is triggered with every property check.

### LINKED ISSUE #439: infinite recursion in systemd units handling
I have noticed that there is infinite recursion when the systemd unit is not installed.

If the systemd unit doesn't exist, the call `systemdServiceIface.getProperty` at
https://github.com/piggz/harbour-amazfish/blob/0665a8dfdedf4ab5a977126a3c09b8d8c3f84b4a/ui/qml/harbour-amazfish.qml#L116-L133
creates a temporary systemd unit, which triggers
https://github.com/piggz/harbour-amazfish/blob/0665a8dfdedf4ab5a977126a3c09b8d8c3f84b4a/ui/qml/harbour-amazfish.qml#L148-L152

This results in a loop that causes the issue.

After creating and removing the systemd unit in ~/.config/systemd/user/harbour-amazfish.service and running systemctl daemon-reload --user, the systemd behavior changes, and I need to reboot to reproduce the issue again.

I am not sure how to check if the interface is valid without creating temporary systemd unit.

jmlich:
If the systemd unit doesn't exist, the call systemdServiceIface.getProperty creates a temporary systemd unit, which triggers onUnitNew. This results in a loop that causes the issue.

Fixes #439

### DIFF
--- a/ui/qml/harbour-amazfish.qml
+++ b/ui/qml/harbour-amazfish.qml
@@ -116,19 +116,9 @@ ApplicationWindowPL
         function updateProperties() {
             if (ENABLE_SYSTEMD === "YES"){
                 var activeProperty = systemdServiceIface.getProperty("ActiveState");
-                if (activeProperty === "active") {
-                    serviceActiveState = true;
-                } else {
-                    serviceActiveState = false;
-                }
-
+                serviceActiveState = (activeProperty === "active");
                 var serviceEnabledProperty = systemdServiceIface.getProperty("UnitFileState");
-                if (serviceEnabledProperty === "enabled") {
-                    serviceEnabledState = true;
-                }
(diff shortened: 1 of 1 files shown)


ui/qml/harbour-amazfish.qml
```diff
@@ -147,7 +137,8 @@ ApplicationWindowPL
         signal unitNew(string name)
         onUnitNew: {
```
piggz, line 138: Should we just remove this entire block then?

jmlich: You are right, looks much better when whole block is removed.

I added just comment because the checking onUnitNew seems to be reasonable, but leads to unexpected infinite loop.

PR #441, opened 2024-12-27
Add translations to huami items

jmlich:
I have noticed that some items are not translated:

screenshot20241227_093249365


jmlich:
I have added all strings from mibandservice.cpp. I am not sure if qml with "const" variable in javascript will work on SailfishOS

jmlich:
I have check this with SailfishOS and it works as expected

Issue #375, opened 2024-05-12
Amazfit NEO pairing


jmlich:
There are many differences from my previous try:
- compiled natively on Fedora 40 as Kirigami flavor
- it is different bluetooth device

I am trying to pair with Amazfit NEO. The device is connected to zepp application and I am getting token with

```
$ python3 -m huami_token -m amazfit -e user@domain.com -p 'Password' -b
...CUT...
╓───Device 0
║  MAC: E1:87:29:30:E4:E7, active: Yes
║  Key: 0xcaa04207ff084f260d8513a5d75d8dcf
╙────────────
```

The `0xcaa04207ff084f260d8513a5d75d8dcf` (including 0x) was entered into dialog. 

I can see authfailed message in daemon's log:

```
2024-05-12 17:34:36.615 : void MiBand2Service::characteristicChanged(const QString&, const QByteArray&) Changed: "00000009-0000-3512-2118-0009af100700" "\x10\x83\b"
2024-05-12 17:34:36.615 : Unexpected data
2024-05-12 17:34:36.615 : void HuamiDevice::authenticated(bool) false
2024-05-12 17:34:36.615 : void AbstractDevice::setConnectionState(const QString&) Connection state: "authfailed"
```

Whole log follows:

```
2024-05-12 17:34:36.526 : 9 nodes
2024-05-12 17:34:36.527 : Creating service for:  "00001801-0000-1000-8000-00805f9b34fb"
2024-05-12 17:34:36.527 : Creating service for:  "0000180a-0000-1000-8000-00805f9b34fb"
2024-05-12 17:34:36.527 : Creating service for:  "00001530-0000-3512-2118-0009af100700"
2024-05-12 17:34:36.528 : Creating service for:  "00001811-0000-1000-8000-00805f9b34fb"
2024-05-12 17:34:36.528 : Creating service for:  "00001802-0000-1000-8000-00805f9b34fb"
2024-05-12 17:34:36.528 : Creating service for:  "0000fee0-0000-1000-8000-00805f9b34fb"
2024-05-12 17:34:36.528 : Creating service for:  "0000fee1-0000-1000-8000-00805f9b34fb"
2024-05-12 17:34:36.529 : Creating service for:  "0000180d-0000-1000-8000-00805f9b34fb"
2024-05-12 17:34:36.529 : Got MiBand2 service
2024-05-12 17:34:36.529 : Starting notify for  "00000009-0000-3512-2118-0009af100700"
[...]

PR #442, opened 2024-12-27
[huami] Avoid repeated initialization

jmlich:
Fixes: #375

I am not sure if this is right solution to the issue. However, I am able to connect to Amazfit NEO with this fix.

I was trying to improve debug messages to be able to find source of issue.


piggz:
What is the actual fix here?  Dont initialize until after 60 seconds?  There are quite a few debug changes that could be removed to make the commit simpler

jmlich:
The debug changes were moved to separate pull request.

piggz:
Have tested and confirmed im still able to connect ok to my GTR2, so should be ok to merge.

PR #457, opened 2025-01-18
[SportSummaryPage] Use sharead Image.source to accelerate rendering of Icons and use Loader

jmlich:
According to my measurement on desktop the loading of Page changed from 1.6 s to 0.1 s after first load of data. It seams there is still some ugly leak which is causing delay in rendering of the page.

I was testing that just with kirigami right now.

Details are shown here: https://youtu.be/wfLNFML_pUA


jmlich:
It seems 
```
    onPageStatusActive: {
        pushAttached(Qt.resolvedUrl("SportsSummaryPage.qml"))
    }
```
Adds new instance on every reload. This should be done just once.

PR #464, opened 2025-01-24
Separate ray and firefish device names in PairSelectDeviceType.qml

eLtMosen:
Attempt to address an issue where the ray and firefish devices do not appear in the pairing UI of Amazfish. It appears that ray and firefish were grouped together in PairSelectDeviceType.qml, whereas they are handled separately in devicefactory.cpp.

This change separates the two devices in the UI, aligning with the handling observed in devicefactory.cpp. While untested, this adjustment aims to resolve the potential mismatch and allow proper pairing for ray and firefish.


jmlich:
Great that you test all AsteroidOS devices.
I have updated the PairSelectDeviceType recently in https://github.com/piggz/harbour-amazfish/pull/458. I have included your fix also into my pull request.

PR #479, opened 2025-03-28
[Pebble] early steps of implementation

jmlich:
This is a very early stage of implementation. Rockwork does not use BLE but instead relies on a Bluetooth socket, which is incompatible with qble in Amazfish. It seems that the BLE implementation in GadgetBridge depends on BLE services on both the client and server, which is also currently incompatible with qble/Amazfish. I encountered the same issue in #341.

Amazfish no longer displays DeviceListModel directly, so the user will not see the Pebble icon immediately upon startup.

Currently only connection and Device Info Service works (i.e. DebugPage shows some details about device)

image
image


norayr:
omg, thank you so much!
i am surrounded by 5 dedicated pebble users, and i myself need its support in amazfish, because i can just use normal linux mobiles other than sailfish to talk with my pebble classic.
i understand pebble 2 uses BLE and that's more complicated, but it would be great to support it as well. here we have two of them, and those are only usable with androids.

PR #488, opened 2025-04-12
Ubuntu touch user metrics

jmlich:
The goal is to display the message `%1 steps made today` on the lock screen, similar to the example in the UBports documentation:
https://docs.ubports.com/de/latest/appdev/guides/user-metrics.html

screenshot20250412_111342762

I suspect there may be an issue with internationalization (i18n), since Amazfish uses lupdate and lrelease, while other UBports apps use gettext. The libuserinput library tries to set the textDomain() to load translations independently.

### DIFF
--- a/click/harbour-amazfish.apparmor
+++ b/click/harbour-amazfish.apparmor
@@ -1,5 +1,6 @@
 {
     "policy_groups": [
+        "usermetrics"
     ],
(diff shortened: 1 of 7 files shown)


piggz:
Wondering if I can add similar to the lockscreen on sailfish

Issue #482, opened 2025-03-30
Dismissing notifications on phone does not dismiss them on the watch


roundedrectangle:
I'm using AsteroidOS, but it seems like it is not supported on any device. I checked libwatchfish and it does seem to include this, but it is not used in amazfish.

PR #489, opened 2025-04-24
Refactor notifications.

piggz:
Allow richer notification alerts by refactoring alert notifications to pass around a WatchNotification object instead of just 3 strings.

This will allow devices which support it, to track notifications by Id, and utilise the application name/id.

There should be no change for older implementations.

### DIFF
--- a/daemon/src/operations/huamiinitoperation2021.cpp
+++ b/daemon/src/operations/huamiinitoperation2021.cpp
@@ -171,9 +171,13 @@ void HuamiInitOperation2021::generateKeyPair()
 void HuamiInitOperation2021::debugArrayPrint(const QString &name, uint8_t *arr, int size)
 {
     qDebug() << name << ":";
+#if defined(MER_EDITION_SAILFISH) || defined(UUITK_EDITION)
+    qDebug() << "FIXME";
+#else
     QDebug dbg(QtDebugMsg);
     for (int i = 0; i < size; i++) {
         dbg << Qt::hex << arr[i] << (i < size ? ':' : '\n');
     }
+#endif
 }
 


daemon/src/devices/asteroidosdevice.cpp
```diff
@@ -33,13 +33,13 @@ AbstractFirmwareInfo *AsteroidOSDevice::firmwareInfo(const QByteArray &bytes)
...
 {
-    qDebug() << Q_FUNC_INFO << sender << subject << message;
+    qDebug() << Q_FUNC_INFO;
 
-    AsteroidNotificationService *notification = qobject_cast<AsteroidNotificationService*>(service(AsteroidNotificationService::UUID_SERVICE_NOTIFICATION));
-    if (notification) {
-        notification->sendAlert(sender, subject, message);
+    AsteroidNotificationService *n = qobject_cast<AsteroidNotificationService*>(service(AsteroidNotificationService::UUID_SERVICE_NOTIFICATION));
+    if (n) {
+        n->sendAlert(notification.appName, notification.summary, notification.body);
```
jmlich, line 42: this should be probably `n->sendAlert(notification);`. In fact, the AsteroidNoificationService might utilize some other details such as notification id internally. If I remember correctly it creates some fake id to pass something. I will try to append patch for that.

jmlich (approved):
Most of it looks good for me. I haven't tested it yet.

daemon/src/devices/asteroidosdevice.cpp
jmlich: This is also related to https://github.com/piggz/harbour-amazfish/issues/482

piggz:
@jmlich i have made some changes to asteroid notification service, but they are untested.  Will merge and let you try this out.

PR #491, opened 2025-04-25
Move WatchNotification into amazfish.h, add conversion to QVariantMap

jmlich:
The WatchNotification struct was moved into `lib/src/amazfish.h` to be able to reference that from UI. Additionally, I have created toQVariantMap fromQVariantMap conversions to be able pass it via dbus.

The interface function was overloaded to be able use it via dbus:
```
void DeviceInterface::sendAlert(const QVariantMap &notification, bool allowDuplicate)
void DeviceInterface::sendAlert(const Amazfish::WatchNotification &notification, bool allowDuplicate)
```

### DIFF
--- a/daemon/src/deviceinterface.cpp
+++ b/daemon/src/deviceinterface.cpp
@@ -183,7 +183,7 @@ HRMService *DeviceInterface::hrmService() const
 
 void DeviceInterface::onNotification(watchfish::Notification *notification)
 {
-    AbstractDevice::WatchNotification n;
+    Amazfish::WatchNotification n;
     n.id = notification->id();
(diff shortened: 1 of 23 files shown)


piggz:
Looks good, just needs a conflict resolved. thx

jmlich:
I will solve that conflict.

I'm still not very happy that the DBus interface doesn't work directly with the WatchNotification class. Removing Q_INVOKABLE from that method might solve the problem, but I'm not sure whether it's used elsewhere. It seems there might be some code that tells DBus how to serialize the object, which could potentially be used in the UI process the same way as in the daemon.

PR #501, opened 2025-05-19
GTR 3 Pro support

b100dian:


ui/qml/components/DeviceListModel.qml
```diff
@@ -48,6 +48,13 @@ ListModel {
         pattern: "Amazfit GTS 2"
     }
 
+    ListElement {
+        deviceType: "Amazfit GTR 3 Pro"
+        icon: "../pics/devices/miband4.png"
+        auth: true
+        pattern: "Amazfit GTR 3 Pro"
+    }
+
```
b100dian, line 57: @piggz I'm pretty sure I lost much time because the "Amazfit GTR" pattern was tested first :-S

Would it mean that the GTR 2 pairing doesn't work well either?

jmlich: You are right, patterns should be updated to avoid confusion between similar devices. Most probably is enough to add ^ to the beginning and $ to the end of expression

jmlich:
It would be great to add also pull request into https://github.com/amazfish/amazfish.github.io/tree/main/_devices to see which features actually works.

ui/qml/components/DeviceListModel.qml
b100dian: True, but maybe somebody relies on this matching e.g. GTR 2 mini or something.. That's why I will just re-order a couple for now. Or maybe we can sort them by length before testing

jmlich: Well, that should be solved by extending pattern.
In 2.6.0 there were exact matching for almost everything except of Bangle.js https://github.com/piggz/harbour-amazfish/blob/2.6.0/daemon/src/devicefactory.cpp#L20 Other way of matching leads to unexpected behaviour.

Pairing with smartwatch outside of pattern should be done by manual selection of device. This is in my todo list, because iband (implemented in "DK08 device") is have large list of devices and it will be easier to choose one.

b100dian: > Well, that should be solved by extending pattern.

Of course, but I don't know _if_ someone relies on a such case. But since you say that "In 2.6.0 there were exact matching for almost everything except of Bangle.js" then probably it's safe that you added $s and ^s :+1:

PR #517, opened 2025-08-01
fix usage of anchors in layout

Danfro:
On FirstPage I get some errors in the log about using anchors in a layout component. Trying to fix this.

I am afraid I can't provide a screenshot or test this in UI since I never see this part. Not sure how to trigger it.


jmlich:
I will try to check if it doesn't break SailfishOS flavor. I am not sure if Layout.alignment is available in Qt 5.6.

jmlich:
Okay, it looks good on SailfishOS

PR #533, opened 2025-11-29
Layout fixes

Danfro:
This PR will address a few layout issues.

1. add missing updown icon for UITK
2. slighly downsize icons for UITK
3. add column spacing for FormLayoutPL

1 fixes that there is no visual indication that the elements in Huami Display Items can be dragged for reordering, although I personally would prefer the icons to be smaller, but this change would then be for all platforms, right?

2 I think icons are much bigger in this app than in other apps on UT, so I think reducing the size a bit is fine. This will be UT only, changing them in other places would affect other platforms too, which I can't test.

3 will fix the entries in device settings page being squeezed together, we could of course use medium spacing too if that would be prefered, but since that is the only place where this is used and there is only a few entries, the large spacing seems fine to me

1 - before (the icon is also missing with the latest PR merged)

grafik

1 - after

grafik

2 - before (example start screen)

grafik

2 - after

grafik

3 - before

grafik

3 - after

grafik

### DIFF
--- a/ui/qml/components/Alarm.qml
+++ b/ui/qml/components/Alarm.qml
@@ -36,7 +36,7 @@ Item {
             }
             ButtonPL {
                 id: button
-                text: "Time: " + pad(hour, 2) + ":" + pad(minute, 2)
+                text: qsTr("Time: %1:%2").arg(pad(hour, 2)).arg(pad(minute, 2))
 
                 onClicked: {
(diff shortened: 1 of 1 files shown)


Danfro:
Just added another small improvement. The button text Time: in Alarm page is not translatable. I made the full string translatable using .arg, because I don't know if some languages would need to shuffle items around.

grafik

PR #536, opened 2025-12-07
rename profile settings to user settings

Danfro:
I personally find it slightly confusing to read "user settings" in the settings menu, but the title of the page actually reads "profile settings". So I suggest to use the same name in both places. Could of course also be profile setting in both places.


jmlich (approved):

Danfro:
@jmlich would I need to edit the file name [here](https://github.com/piggz/harbour-amazfish/blob/master/ui/harbour-amazfish-ui.appdata.xml#L43) too? Or is that file somehow autogenerated?

I hope I found all other spots and renamed the files to match the new string.

jmlich:
The appdata XML is maintained manually. We need to update it if we rename a screenshot. I suggest keeping the screenshot names the same. I don’t remember how to configure Weblate to use different screenshots.

PR #537, opened 2025-12-07
adjust regex to allow the gtr2e

Danfro:
With the current regex the gtr2e is not matched. This PR allowes the gtr2e to be matched as well.


Danfro:
From what I read, the question mark makes the character before it optional. [regex101.com](https://regex101.com/) explains it like this "? matches the previous token between zero and one times". When testing there it works. Without question mark it won't work. But I admit, to me regex is a bit like black magic in coding world. :grinning:

piggz:
Ah ok, makes sense then!

PR #545, opened 2026-01-11
reduce logging

Danfro:
I think sometimes we do log too much data. Some is simply not needed, some may be considered a privacy issue. And I am not sure if they are all helpful when troubleshooting.

Please see this rather as suggestion. I don't have that much experience with troubleshooting, so some might have proofed useful in the past. But some may be leftovers... I know some are from my own doing. :-)

### DIFF
--- a/daemon/src/deviceinterface.cpp
+++ b/daemon/src/deviceinterface.cpp
@@ -102,7 +102,7 @@ DeviceInterface::~DeviceInterface()
 
 void DeviceInterface::connectToDevice(const QString &address)
 {
-    qDebug() << Q_FUNC_INFO << ": address:" << address;
+    // qDebug() << Q_FUNC_INFO << ": address:" << address;
 
 
     if (m_device) {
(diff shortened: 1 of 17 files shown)


piggz:
Probably fine, but id prefer to delete the code rather than comment out :)

Danfro:
> Probably fine, but id prefer to delete the code rather than comment out :)

There are already some commented out log commands in the code. And I am not sure if we want to keep them for quicker logging in case we need it for debugging. But some I can certainly delete. If you are happy with all of them being deleted, I can of course do that too.

daemon/src/deviceinterface.cpp
```diff
@@ -102,7 +102,7 @@ DeviceInterface::~DeviceInterface()
 
 void DeviceInterface::connectToDevice(const QString &address)
 {
-    qDebug() << Q_FUNC_INFO << ": address:" << address;
+    // qDebug() << Q_FUNC_INFO << ": address:" << address;
```
jmlich: I would prefer to keep this one.

ui/src/trackloader.cpp
```diff
@@ -139,7 +139,7 @@ void TrackLoader::vWriteFile(const QString &sFilename)
     }
     else
     {
-        qDebug() << "error opening output file\n";
+        qDebug() << "error opening output file\n"<<fullFilename;
```
jmlich: This fix is great, but I would remove \n

Issue #548, opened 2026-01-17
CREATE INDEX ON mi_band_activity (timestamp);


jmlich:
I have done some simple performance test with PinePhone. 

I suggest to create following index 
```
CREATE INDEX idx_mi_band_activity_timestamp ON mi_band_activity (timestamp);
```
I am not sure what is the best place to create index and how can we make sure that existing databases are updated.

Follows performance test. Loading of attached page Steps:

- before creating index (3 seconds):
```
Jan 17 17:04:13 PinePhone i[6062]: [D] unknown:0 - "" QQuickGridLayoutBase::rearrange() QSizeF(720, 70)
Jan 17 17:04:13 PinePhone i[6062]: [D] DataSource::data:179 - "SELECT date(timestamp_dt), sum(steps) FROM mi_band_activity WHERE date(timestamp_dt) >= date('2026-01-17T00:00:00','-10 day') AND timestamp_dt <= '2026-01-17T23:59:59' GROUP BY date(timestamp_dt) ORDER BY timestamp_dt ASC"
Jan 17 17:04:14 PinePhone i[6062]: [D] unknown:0 - QQuickGridLayoutBase::invalidate()
Jan 17 17:04:14 PinePhone i[6062]: [D] unknown:0 - engine::invalidate()
Jan 17 17:04:14 PinePhone i[6062]: [D] unknown:0 - QQuickLayout::invalidate(), polish()
Jan 17 17:04:14 PinePhone i[6062]: [D] DataSource::data:179 - "SELECT date(timestamp_dt), sum(steps) FROM mi_band_activity WHERE date(timestamp_dt) >= date('2026-01-17T00:00:00','-10 day') AND timestamp_dt <= '2026-01-17T23:59:59' GROUP BY date(timestamp_dt) ORDER BY timestamp_dt ASC"
Jan 17 17:04:16 PinePhone i[6062]: [D] unknown:0 - "" QQuickGridLayoutBase::rearrange() QSizeF(720, 70)
```

- after creating index (2 seconds):
```
Jan 17 17:05:24 PinePhone i[6093]: [D] unknown:0 - "" QQuickGridLayoutBase::rearrange() QSizeF(720, 70)
[...]

PR #555, opened 2026-02-01
CREATE INDEX ON mi_band_activity (timestamp);

jmlich:
Fixes: #548

I’m not sure if this is the correct way to create an index for a Kexi database.

### DIFF
--- a/daemon/src/deviceinterface.cpp
+++ b/daemon/src/deviceinterface.cpp
@@ -266,6 +266,10 @@ void DeviceInterface::createTables()
         qDebug() << *t_activity;
     }
 
+    if (!m_conn->executeSql(KDbEscapedString("CREATE INDEX IF NOT EXISTS idx_mi_band_activity_timestamp ON mi_band_activity (timestamp);"))) {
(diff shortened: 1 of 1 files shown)


daemon/src/deviceinterface.cpp
```diff
@@ -266,6 +266,10 @@ void DeviceInterface::createTables()
         qDebug() << *t_activity;
     }
 
+    if (!m_conn->executeSql(KDbEscapedString("CREATE INDEX IF NOT EXISTS idx_mi_band_activity_timestamp ON mi_band_activity (timestamp);"))) {
```
piggz: I wonder if you just need to add KDbField::Indexed to the field type for the timestamp field?

jmlich: probably yes, but I wasn't able to figure out syntax for that.

jmlich: Feel free to fix it as you suggest. I will get back to this soon (I hope) and will propose some update.

piggz:
Have implemented slightly differently, but kept the same create index code

jmlich:
:heart: Great!

PR #556, opened 2026-02-01
Show page when feature is supported

jmlich:
TODO: reload when supported features changed after reconnected or new pairing

### DIFF
--- a/ui/qml/harbour-amazfish.qml
+++ b/ui/qml/harbour-amazfish.qml
@@ -11,7 +11,7 @@ import "./components/platform"
 ApplicationWindowPL
 {
     id: app
-    initialPage: Component { FirstPage { } }
+    initialPage: Qt.createComponent( Qt.resolvedUrl("pages/"+pagesModel.get(0).qmlUrl) )
     property var    rootPage: null
     cover: Qt.resolvedUrl("cover/CoverPage.qml")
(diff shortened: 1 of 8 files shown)


jmlich:
I am still not convinced this is a right way. 

While I am considering using [SwipeView](https://doc.qt.io/qt-6/qml-qtquick-controls-swipeview.html) instead of pagestack to reach same users experience of SFOS and other platforms. It would be great also to improve ux on desktop. The platform.{kirigami,uuitk,sfos} layer is causing headache for those drafts. 

I thing dialog list shouldn't change even when other device paired. As far as you have some data stored you may want to look on them.

I will probably try to work on this prototype further to see how it will end up.

Thaodan:
Personally I hope the page stack stays but the pulley menu on the initial page has to many submenus that don't belong there. I.e. pair with watch shouldn't be there maybe when you are already connected or at least go to the settings page after the initial pairing? The other option to view data graphs should go to the stats page.

Install file could go to the settings page.

jmlich:
The page stack itself works nicely with the SailfishOS UI right now, but the Kirigami implementation doesn’t look as I would expect in full-screen mode:

image

Anyway, my conclusion is that I need to go back to the whiteboard or Figma and properly design what I want—and make it flexible enough to support multiple devices, with the ability to hide features that aren’t available on specific devices.

PR #557, opened 2026-02-05
Drop pattern in ci

jmlich:


jmlich:
It should fix this
image

PR #598, opened 2026-03-22
Use amber-mpris instead of qtmpris in CI

jmlich:
With https://github.com/piggz/libwatchfish/pull/18 we need to update packaging/compilation on all platforms
- [x] ubuntu touch - clickable
- [x] SailfishOS - rpm
- [ ] flatpak (not in amazfish repo)
- [ ] alpinelinux (not in amazfish repo)

Effectively, merge of updated libwatchfish will break flatpak ci until new version release. I am not sure if we want to merge now. I would consider to merge it shortly before release. On the other hand it means will be not tested on with other ci jobs.


jmlich:
I mean, the change is ready. The flatpak is hard to update in CI and it doesn't make much sense as it must be reworked after release again. I have tried flatpak version and it worked.
We can wait with merge until shortly before release to keep whole CI working as long as possible

PR #600, opened 2026-03-23
New UI

piggz:
Create a new UI adding much more functionality to the main page.

Tile based UI gives easy access to any other page without swiping through several pages.

Tiles can show additional info, this is implemented in the steps, pai, hr and battery tiles.

Tiles are one shown if a device supports them.


piggz:
image

Sailfish UI.   Some minor tweaks needed to some size/position of the HR action or text, but not bad.

PR #603, opened 2026-03-29
Fixes segfault caused by missing clear in reset() in triggerNextOperation()

jmlich:
The segfault was reported in #601; this change does not fix the issue itself.

I’m not able to verify whether this fix helps. @pemekcz, could you please test it?

RPM files should be available for download in the artifacts: https://github.com/piggz/harbour-amazfish/pull/603/checks

According to the example at https://doc.qt.io/qt-6/qtalgorithms.html#qDeleteAll, `qDeleteAll` is followed by `clear()`. I believe this is the correct way to use it. I just don’t know how to verify the scenario in which the issue was reported.

### DIFF
--- a/daemon/src/devices/huami/huamifetcher.cpp
+++ b/daemon/src/devices/huami/huamifetcher.cpp
@@ -105,6 +105,7 @@ void HuamiFetcher::reset()
         m_device->message(tr("All operations cancelled"));
     }
     qDeleteAll(m_operations);
+    m_operations.clear();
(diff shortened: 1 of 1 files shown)


pemekcz:
Fix verified the same way I reproduced it. 

Repro steps:
- disconnect BT device, reconnect without authentication
- in settings/debug info, use Fetch data button with eg 1023 mask few times, requesting data faster than respomse time.
- wait until response came. Should not crash.

PR #608, opened 2026-04-04
Part of condition always return true (_authenticated || !_authenticated)

jmlich:


piggz:
there is actually an error here, the trick should use && instead, so should be (_authenticated && !_authenticated) ... such that any change to authentication re-evaluates the binding.  see https://stackoverflow.com/questions/48325115/qml-programmatically-update-binding

PR #622, opened 2026-04-19
hide map when no coord is present and elide label column

Danfro:
I think it makes no sense to display the map when e.g. an indoor sport has no coordinate available. So hide it.

Also the left column texts can be quite long when translated. So better elide them to avoid overlapping to the second column.


Danfro:
Although we may also consider having different width of the two data columns. The values on the right side will generally be shorter than the labels on the left.

jmlich:
lgtm. I have realised that the detection of invalid position didn't worked for me properly. I have fixed it in the next commit.

PR #627, opened 2026-05-11
Add option to filter out email app notifications to watch

oles1234:
My watch was going crazy with notifications every time I checked my email with the stock SFOS Email app (I get a lot of emails), and there was no way to disable/filter the notifications on new emails anywhere. Several users also asked about this on the SFOS forum, but there was no solution. This patch adds an option to the Amazfish Application Settings menu "No notifications for emails" that does exactly that. Default is off, i.e. no change. No other side effects.

With the new option on (and saved!), notifications with appID "messageserver5" or appName "Mail" will be ignored and not sent to the watch. This covers exactly the notifications from the Mail app.

### DIFF
--- a/daemon/src/deviceinterface.cpp
+++ b/daemon/src/deviceinterface.cpp
@@ -178,6 +178,14 @@ HRMService *DeviceInterface::hrmService() const
 
 void DeviceInterface::onNotification(watchfish::Notification *notification)
 {
+    if ((AmazfishConfig::instance()->appEmailsNotification()) &&
+     (notification->appId() == "messageserver5" ||
+      notification->appName() == "Mail"))
+    {
+        qDebug() << "Ignoring Sailfish Mail notification";
+        return;
+    }
+
     Amazfish::WatchNotification n;
     n.id = notification->id();
     n.appId = notification->appId();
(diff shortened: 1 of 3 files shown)


jmlich:
Great idea!

It would be nice to be able to select which notifications to filter.
For example, here I manually removed `ayatana-indicator-sound`:
https://github.com/piggz/libwatchfish/blob/41b885d5191881379200882ffe76b1709b559fdb/notificationmonitor.cpp#L139

It would be great to have some more generic way to do this.

I can see you added this condition for filtering:

```cpp
(notification->appId() == "messageserver5" ||
 notification->appName() == "Mail")
```

Isn't `appId()` enough?

Some time ago I was trying to collect a list of `appId`s so I could assign the correct icon:

```sh
journalctl | grep harbour-amazfish | grep Notif | grep appId | sed 's/.*appId="//' | sed 's/".*//' | sort | uniq
```

oles1234:
I guess appId() alone would suffice, I just didn't want to experiment too much at the time. As a more generic way to do it I would try to populate a list of seen appId/appName at runtime, in the same place as the filtering is done, and let the user choose to filter each by name in a submenu. I guess the appName isn't necessarily unique but more meaningful than appId alone so it helps the user find what they really want to block. It's a pragmatic approach, but at the same time also convenient because only apps that actually give a reason to be filtered will show up.

jmlich:
Please take a look on https://github.com/piggz/harbour-amazfish/pull/628
I have created draft of this with settings. I will be happy for comments and for test with sailfish os.
It needs some work still.

jmlich:
Closing as this should be fixed by #628

PR #632, opened 2026-05-26
[Garmin] Initial Support

thmichel:
Realtime HR and Steps support for Garmin Epix Pro, should work with any Garmin device


daemon/src/devices/garmin/garmindevice.cpp
```diff
@@ -0,0 +1,244 @@
+#include "garmindevice.h"
+#include "services/garmin/communicator_v2.h"
+#include "hrmservice.h"
+#include "deviceinfoservice.h"
+#include "amazfishconfig.h"
+
+
+#include <QDomDocument>
+#include "MNotification"
```
jmlich: I think this include should have been removed in 83ff2f489cf1bd851033af5b0d10630bf50600eb.

thmichel: That's right, forgot to remove that. I'm working onfurther integrations now, will be removed with one of the next commits. Feel free to remove it in upstream already, don't think it makes sense to create an extra pull request for this.

daemon/src/devices/garmin/garmindevice.cpp
```diff
@@ -0,0 +1,244 @@
...
+ }
+
+
+void GarminDevice::onPropertiesChanged(QString interface, QVariantMap map, QStringList list)
+{
+    qDebug() << Q_FUNC_INFO << interface << map << list;
+
+    if (interface == "org.bluez.Device1") {
+        m_reconnectTimer->start();
+        if (deviceProperty("ServicesResolved").toBool() ) {
```
jmlich, line 128: I think this one should look like this:
```
        if (map.contains("ServicesResolved") && map["ServicesResolved"].toBool()) {
            initialise();
        }
```

daemon/src/devices/garmin/garmindevice.cpp
```diff
@@ -0,0 +1,244 @@
...
+    qDebug() << Q_FUNC_INFO << "Pairing with Garmin Epix";
+
+    m_needsAuth = true;
+    m_pairing = true;
+    m_autoreconnect = true;
+    setConnectionState("pairing");
+    emit connectionStateChanged();
+
+    QBLEDevice::pair();
+    initialise();
```
jmlich: I believe the flow should be 
(1) constructor GarminDevice() makes connect propertiesChanged
(2) QBLEDevice::pair(); causes change of properties
(3) onPropertiesChanged calls
if (map["ServicesResolved"].toBool()) {
  (4) initialize();
}
Then services should be created.

Calling initialize here looks too early. Can you check if it works for you without initialise() on this line?

daemon/src/devices/garmin/garmindevice.cpp
```diff
@@ -0,0 +1,244 @@
...
+    QDBusReply<QString> xml = adapterIntro.call("Introspect");
+
+
+    QDomDocument doc;
+    doc.setContent(xml.value());
+
+    QDomNodeList nodes = doc.elementsByTagName("node");
+
+    qDebug() << nodes.count() << "nodes";
+
```
jmlich, line 176: I am using 
    qDebug().noquote() << doc.toString(4);
here to get some debug information about services

jmlich (commented):
I am not able to test with Instinct Solar. The ServicesResolved stays false. I am not sure what to change.

daemon/src/devices/garmin/garmindevice.cpp
thmichel: I'll check as soon as I got my installation issue sorted - my device updated to 5.1, now I get an error when installing my rpm:
	libicui18n.so.70()(64bit) wird benötigt von harbour-amazfish-2.9.0-1.aarch64
	libicuuc.so.70()(64bit) wird benötigt von harbour-amazfish-2.9.0-1.aarch64

thmichel: Got the app working again, but can't pair a device any more, get the message "Method 'Pair' with Signature ''  on interface 'org.bluez.Device1' doesn't exist". 
Currently trying to find out if the 5.1 update broke it or something I did. Strange that even with the Amazfish app from Chum I can't connect to any bluetooth device any more, always get this eror message. Too bad I didn't notice I was still in the EA program. Need some time to investigate.

jmlich: I was trying to improve pairing process in https://github.com/piggz/harbour-amazfish/pull/522, but there was some timeout issue which needs more investigation. I recommend to check device in system wide bluetooth settings, remove everything, reapply  [Appearance](https://amazfish.github.io/Appearance) if needed settings and reboot. The amazfish will show same error message during unpairing. It might be necessary to set device trusted manually (in system wide bluetooth settings). Additionally, pairing process might differ on every platform.

thmichel: Found the issue - in 5.1, you cann add bluetooth support for AAS via an app (which I did). This seems to change the adapter path.  Amazfish shows /org/bluez/hci1 in the adapter settings then, while nthe watch address contains hci0. After removing this app, everything works again.  I did remove the "emit connectionstatechanged" and "initialize" from the pair function and can pair the watch now again.  The message "No Garmin device detected" pops up, but it works. I need to check why that happens, seems parseSrvices is called multiple times and I don't check for an existing service first.

We should also investigate into the hci1 issue, I think that many users will enable bluetooth support for AAS in 5.1 as it makes lots of android applications work that use bluetooth connections.

piggz: Ah yes, I got warned about this from Jolla, need to add a todo for it!  we could store just the device mac, then search for it on the dbus path and ignore the hci device?

thmichel: All work now as expected again. I can connect to the Epix Pro without error messages, also connection to a VivoSmart 5 works, but unreliably. Could be because it doesn't support all features I use, I think it makes sense to create a general Garmin device with minimum features and then derive the differen watches from that, similar to the abstractdevice model. For the hci1 issue, I'll raise a new issue so it can be properly tracked.

ui/qml/components/DeviceListModel.qml
```diff
@@ -135,5 +135,16 @@ ListModel {
...
     }
-
+    ListElement {
+        deviceType: "Generic Garmin"
+        icon: "../pics/devices/kospet_dk08.png"
+        auth: false
+        pattern: "GARMIN"
+    }
+    ListElement {
+        deviceType: "Garmin Epix"
```
jmlich, line 145: deviceType String must match with deviceMap record at daemon/src/devicefactory.cppL50
`s/Garmin Epix/Garmin Epix Pro/`

ui/qml/pages/Spo2DataPage.qml
```diff
@@ -30,10 +30,9 @@ PagePL {
             font.pixelSize: styler.themeFontSizeExtraLarge * 3
             anchors.horizontalCenter: parent.horizontalCenter
             width: parent.width
-            text: graphSpo2Normal.lastValue + "%"
+            text: supportsData(Amazfish.TYPE_SPO2) ? qsTr("%1%").arg(_InfoSpo2) : graphSpo2Normal.lastValue + "%"
```
jmlich, line 33: I suggest wrapping the condition other way around:
```
            text: qsTr("%1%").arg( supportsData(Amazfish.TYPE_SPO2) ? _InfoSpo2 : graphSpo2Normal.lastValue )
```

daemon/src/services/garmin/garmintypes.h
```diff
@@ -0,0 +1,192 @@
...
+quint32 le32(const char* p);
+quint64 le64(const char* p);
+quint8  u8le(const QByteArray& b, int off);
+quint16 u16le(const QByteArray& b, int off);
+quint32 u32le(const QByteArray& b, int off);
+quint64 u64le(const QByteArray& b, int off);
+qint32  i32le(const QByteArray& b, int off);
+void writeU16le(QByteArray& out, quint16 v);
+void writeU32le(QByteArray& out, quint32 v);
+void writeU64le(QByteArray& out, quint64 v);
```
jmlich, line 291: Looks like, there are already similar functions in TypeConversion

daemon/src/services/garmin/garmintypes.cpp
```diff
@@ -0,0 +1,271 @@
...
+}
+
+// =============================================================================
+// Service conversions + Display [1](https://computacenter-my.sharepoint.com/personal/thomas_michel_computacenter_com/Documents/Microsoft%20Copilot%20Chat%20Files/types.rs)
+// =============================================================================
+
+Result<Service> serviceFromCode(quint16 code) {
+    switch (code) {
+    case 1:      return Result<Service>::isOk(Service::GFDI);
+    case 4:      return Result<Service>::isOk(Service::Registration);
```
jmlich, line 131: On first look it looks like `enum class Service : quint16`  contain constants already:
e.g. this line can be rewritten like this:
```
 case Service::Registration return Result<Service>::isOk(Service::Registration);
```
or `whole switch(code) { ... }` can be shrieked to
```
return Result<Service>::isOk(code)
```

It looks like it might be necessary that code is valid value in enum. It would be great to have some elegant solution.

daemon/src/services/garmin/garmintypes.cpp
```diff
@@ -0,0 +1,271 @@
...
+    case 0xA018: return Result<Service>::isOk(Service::FileTransferA);
+    case 0xC018: return Result<Service>::isOk(Service::FileTransferC);
+    case 0xE018: return Result<Service>::isOk(Service::FileTransferE);
+    default:
+        return Result<Service>::err(GarminError::unknownServiceCode(code));
+    }
+}
+
+QString serviceToString(Service s) {
+    switch (s) {
```
jmlich, line 154: In general I would prefer some solution such as
In general I would prefer something like
https://github.com/piggz/harbour-amazfish/blob/52c6fdf15dbf1eeb0b76331411040591cd76af5f/lib/src/activitykind.cpp#L4-10
and here
https://github.com/piggz/harbour-amazfish/blob/52c6fdf15dbf1eeb0b76331411040591cd76af5f/lib/src/activitykind.cpp#L463

Simply some const QMap<Service, QString> and lookup

daemon/src/services/garmin/garminnotificationhandler.h
```diff
@@ -0,0 +1,393 @@
...
+    qint32 notificationId = 0;
+    bool hasActions = false;
+    bool hasPicture = false;
+};
+
+//
+// ---------------------------------------------------------------------
+// NotificationSpec
+// ---------------------------------------------------------------------
+class NotificationSpec
```
jmlich, line 48: It is necessary to create new structure for Notifications? It looks like it overlaps with libwatchfish/Notification.h definition. It might be more efficient to stick to existing structure

daemon/src/services/garmin/garminnotification.h
```diff
@@ -0,0 +1,393 @@
...
+
+
+
+// Helper functions (Rust impl NotificationType)
+struct NotificationTypeUtil
+{
+    static quint8 categoryValue(NotificationType t)
+    {
+        switch (t) {
+        case NotificationType::GenericPhone:       return 1;  // INCOMING_CALL
```
jmlich: Currently I don't see how is mapping from app to notificationType is done. This seems to be additional translation layer. 

Some example mapping can be seen here with bangle.js:
enum:
https://github.com/piggz/harbour-amazfish/blob/52c6fdf15dbf1eeb0b76331411040591cd76af5f/daemon/src/devices/banglejsdevice.h#L23
find value:
https://github.com/piggz/harbour-amazfish/blob/52c6fdf15dbf1eeb0b76331411040591cd76af5f/daemon/src/devices/banglejsdevice.cpp#L99

daemon/src/services/garmin/asyncmessagehandler.cpp
```diff
@@ -0,0 +1,277 @@
...
+    // fill packet size = message.len() + 2
+    const quint16 packetSize = quint16(msg.size() + 2);
+    msg[0] = char(packetSize & 0xFF);
+    msg[1] = char((packetSize >> 8) & 0xFF);
+
+    // crc16
+    const quint16 crc = computeCrc16A001(msg);
+    writeU16le(msg, crc);
+
+    emit logDebug(QStringLiteral("📤 Chunk message size: %1 bytes").arg(msg.size()));
```
jmlich: why not just `qDebug() << `

daemon/src/devices/garmin/garmindevice.cpp
```diff
@@ -0,0 +1,265 @@
...
+void GarminDevice::sendAlert(const Amazfish::WatchNotification &notification)
+{
+    qDebug() << Q_FUNC_INFO;
+    CommunicatorV2 *svc = qobject_cast<CommunicatorV2*> (service(CommunicatorV2::UUID_SERVICE_GARMIN_ML_GFDI));
+    NotificationSpec note;
+    note.body = notification.body;
+    note.sourceName = notification.appName;
+    note.title = notification.summary;
+    note.id=0;
```
jmlich: I was trying to rework notification to be able dismiss notification (other branch, work in progress). Original id can be probably used directly (which is preferred), or some map of garmin notification id and system notification id must be done.

daemon/src/services/garmin/asyncmessagehandler.cpp
```diff
@@ -0,0 +1,277 @@
...
+                  .arg(nextOffset)
+                  .arg(chunkInfo.totalLength)
+                  .arg(remainingSaturating));
+
+    if (nextOffset < chunkInfo.totalLength) {
+        const int remaining = chunkInfo.totalLength - nextOffset;
+        const int chunkSize = qMin(remaining, MAX_PROTOBUF_CHUNK_SIZE);
+
+        if (nextOffset + chunkSize > chunkInfo.completePayload.size()) {
+            emit logError(QStringLiteral("❌ ERROR: Chunk boundary exceeds payload! offset=%1 size=%2 total=%3")
```
jmlich: why not qError() ?

jmlich (commented):
I have briefly looked into code and added some notes. Usually, I am trying to find clean solution or asking for it. In the same time I am happy when code is just working. Improvements can be reached in future. A least some code can bring us more testers.
I will try to provide more feedback yet. I am still not lucky to pair my device.

thmichel:
Most of this code is just the initial translation from Rust and needs complete rework, I will look at your comments. I am currently trying to tidy all this up to get notifications work, also trying to get my head around how to distribute functions between the communicator and the messaage handlers.
Current status is I can send GFDI notification messages to the watch and get acknowledgments, but the watch isn't requesting the notification data.
I shouldn't have included this code in the pull request, as this was just to add the realtime data. 

Need to learn to work with branches to avoid this in the future. I don't have too much experience with git.

jmlich:
I have renamed the pull request because chum for chum. It automatically generates changelogs using brackets. See https://sailfishos-chum.github.io/apps/harbour-amazfish/

thmichel:
I looked at the pairing process again, it was really a bit messy. The necessary pairing messages (Sync complet, Sync ready) where not properly sent to the watch as  completepairing() was called before the messagehandler was ready (and was called multiple times).

I've reworked the order for the pairing and initialization and can now reliably pair both a Vivosmart 5 (as Garmin Generic) and my Epix Pro.

What still doesn't work is sending of notifications - the watch does not send the subscribe message after the sync complete message as documented in the GadgetBridge code. 

I am trying to find out why that is.

daemon/src/services/garmin/garminnotificationhandler.h
thmichel: I will look at this, currently I am trying to get the basic notification done using the Gadgetbridge logic, just to understand how it works. This code is very experimental.

daemon/src/services/garmin/asyncmessagehandler.cpp
thmichel: Done, will be in the next commit.

daemon/src/services/garmin/asyncmessagehandler.cpp
thmichel: Done, will be in the next commit.

.gitmodules
```diff
@@ -1,9 +1,9 @@
-[submodule "qble"]
```
jmlich, line 1: The change in .gitmodules look strange. It looks like it was reorder, but it breaks compilation on my computer when I do `git submodule update --init`.

piggz:
Could we avoid emojis in the source code please.  There seems to be some in some debug logs?

.gitmodules
thmichel: I think that was due to a mistake on my end when merging the upstream changes back, had an issue with the submodule and needed to re-add it. Not sure why that results in compliation errors, works here.

thmichel:
> Could we avoid emojis in the source code please. There seems to be some in some debug logs?

I am removing all of them in the process of restructuring the code. Curretly splitting the code into smaller modules for more readability, like one class per message type.

thmichel:
I'm currently splitting the code into smaller chunks as mentioned above. Adds lots more files but I think it increases code readability and helps enabling feature by feature. Battery Level is now reported correctly into Amazfish and shown in the GUI.

I still can't get notifications to work. The watch acknowledges the notification message but does not request content though the subscription message comes in now and is acknowledged by Amazfish to the watch. 
Anyway, at least some progress getting battery level now.

@jmlich : Are you able to pair the Instinct now with the pairing code reworked?

.gitmodules
jmlich: It looks like CI https://github.com/piggz/harbour-amazfish/actions/runs/27163815313/job/80272393043?pr=632 fails with following message:
```
  add_subdirectory given source "qble" which is not an existing directory.
```
which indicates that qble wasn't cloned properly.

jmlich:
I am not able to compile it right now
```
/home/jmlich/workspace/ubports/harbour-amazfish/daemon/src/services/garmin/garminnotification.cpp: In member function ‘void GarminNotificationHandler::onSetCallState(const CallSpec&)’:
/home/jmlich/workspace/ubports/harbour-amazfish/daemon/src/services/garmin/garminnotification.cpp:148:68: error: class template argument deduction failed:
  148 |         notif.attachedActions.insert(0, QSharedPointer(new Action()));
      |                                                                    ^
/home/jmlich/workspace/ubports/harbour-amazfish/daemon/src/services/garmin/garminnotification.cpp:148:68: error: no matching function for call to ‘QSharedPointer(Action*)’
  • there are 12 candidates
In file included from /usr/include/qt5/QtCore/qsharedpointer.h:48,
                 from /usr/include/qt5/QtCore/qpointer.h:43,
                 from /usr/include/qt5/QtCore/QPointer:1,
                 from /home/jmlich/workspace/ubports/harbour-amazfish/daemon/src/services/garmin/communicator_v2.h:8,
                 from /home/jmlich/workspace/ubports/harbour-amazfish/daemon/src/services/garmin/garminnotification.h:12,
                 from /home/jmlich/workspace/ubports/harbour-amazfish/daemon/src/services/garmin/garminnotification.cpp:1:
```

thmichel:
I have reworked the whole protobuf handling to cater for chunked protobuf messages (needed to download fit data). I have no compiling issues, maybe you can test with latest commits?

piggz:
@thmichel Could you rebase this on master please, and if its working we can look to merge.  I see somone ont he forum used an LLM to implement a daemon for garmin. https://forum.sailfishos.org/t/vapaamin-garmin-notifications-bridge/33097

sznowicki:
> @thmichel Could you rebase this on master please, and if its working we can look to merge. I see somone ont he forum used an LLM to implement a daemon for garmin. https://forum.sailfishos.org/t/vapaamin-garmin-notifications-bridge/33097

that's me, if I can help somehow I'd be more than happy to chip in. Yes all the work I did was with LLM but over those two days I have some understanding now how this works.

thmichel:
Sure, I'll have a look at it - did  a lot of work in the background to clean up code. I'll see how to rebase.
@sznowicki: Sounds great, I'll look at your code and may "steel" some parts of it as I can connect to the watch but haven't yet managed to send notifications - the watch does not subscript to them. Might contact you directly if I need help if that's ok?

sznowicki:
> Sounds great, I'll look at your code and may "steel" some parts of it as I can connect to the watch but haven't yet managed to send notifications - the watch does not subscript to them. Might contact you directly if I need help if that's ok?

From what I learned the notification part is tricky but here's the whole deal:

1. Watch for some reason always starts with subscribe propose: disabled
2. We need to tell it we [want it enabled](https://codefloe.com/sznowicki-sailfish/vapaamin/src/commit/baf924b24472ed5a0f0c806f0f3d56a2fd0e0117/src/Vapaamin/src/bluez/DeviceConnection.cpp#L1049)
3. Then watch agrees and responds with proposed: enabled
4. Then sending also works in steps, [we announce a notification](https://codefloe.com/sznowicki-sailfish/vapaamin/src/commit/baf924b24472ed5a0f0c806f0f3d56a2fd0e0117/src/Vapaamin/src/bluez/DeviceConnection.cpp#L522)
5. And then we expect the watch to agree and later ask for notification payload, and this is where we can send the [actual payload](https://codefloe.com/sznowicki-sailfish/vapaamin/src/commit/baf924b24472ed5a0f0c806f0f3d56a2fd0e0117/src/Vapaamin/src/bluez/DeviceConnection.cpp#L1073).

thmichel:
> From what I learned the notification part is tricky but here's the whole deal:
> 
> 1. Watch for some reason always starts with subscribe propose: disabled
> 2. We need to tell it we [want it enabled](https://codefloe.com/sznowicki-sailfish/vapaamin/src/commit/baf924b24472ed5a0f0c806f0f3d56a2fd0e0117/src/Vapaamin/src/bluez/DeviceConnection.cpp#L1049)
> 3. Then watch agrees and responds with proposed: enabled

From what I see in your code, you call handleNotificationSubscription only as a reply to an incoming subscription request (line 1031). That's what I've coded too, but he issue I have is that the watch never sends that Subscription request so I can't reply to it. According to the original Gadgetbridge documentation, the watch should sent that request after the initial connection. I need to look into this again.

sznowicki:
Isn't it needed to send one of those whatever mumbo jumbo characteristics are happening here?

https://codefloe.com/sznowicki-sailfish/vapaamin/src/commit/baf924b24472ed5a0f0c806f0f3d56a2fd0e0117/src/Vapaamin/src/bluez/DeviceConnection.cpp#L788

thmichel:
Yes, that part is working. I can pair the watch and get the send and receive characteristics. I get realtim HR/steps and battery level. So GFDI in general works. When I send notifications, the watch simply ignores them, I guess as the subscription message hasn't come in. That's the point I'm missing: How to get the watch to send the subscription request.

thmichel:
I am finally able to send test messages to the watch reliably! So hopefully by the end of the week I've got the code working with real notification. THen I just need to find out how to rebase so this code can be merged :-)

thmichel:
@sznowicki : Thanks for your code, it helped me a lot in troubleshooting the issue.

piggz:
@jmlich one this is re-based, can you run your bot for any obvious issues.  Im happy to merge so long as it doesnt break any other device, and I have no way of testing it other than build/review.

thmichel:
I've rebased now, following is working:

- Connect and pair a watch
- Get Realtime HR, Spo2 and Steps
- Get battery level from watch
- Set time on watch to device's time
- Send notfications to watch
- Dismissing notifications on watch

Calls cannot be answered / declined from the watch for now, they are shown as incoming calls with Accept and Decline buttons, but the buttons don't work yet.

Getting historic data or activities is not yet implemented.

The code is still quite redundant and probably overcomplex as I initially used AI to port some parts from the Rust implementation. I've cleaned up a lot but there is still a lot to do to make this less ugly.  But at least it's working with the (for me) most important functions now.

Tested with an Epix Pro Gen 2 and a vivosmart 5.

thmichel:
I've added accepting/rejecting of calls - the watch supports rejecting with sending a reply as well, but that would require Amazfish to be able to send an SMS so I disabled that. Thought this made sense before merging as the watch always shows these buttons but they were not working.
Also enabled support for V0/V1 protocol, though I don't have a device to test. 

I won't push other commits to my repo now to give you time to inspect and merge.

jmlich:
We should create API for SMS into libwatchfish.

thmichel:
Yes, I've tried to make a dbus call to test but failed. Not sure if this is due to access restirction. Would be great to have this so you can quickly reply to calls / messages from the watch.

thmichel:
Something different.... I'm currently doing the calendar integration for a future merge request. I noticed that on the watch the end date is always the start date, I guess due to this line in calendarsource.cpp:

events.append(d->convertToEvent(it.incidence(), it.occurrenceStartDate(), it.occurrenceStartDate())); //!TODO end date is newer api

Changing this to use occurenceEndDate() fixes it. I there any harm in doing so?

jmlich:
I think you should make pull request on libwatchfish with it.occurrenceEndDate())). Personally, I haven't seen this issue because I am using ubuntu touch.

jmlich:
> Yes, I've tried to make a dbus call to test but failed. Not sure if this is due to access restirction. Would be great to have this so you can quickly reply to calls / messages from the watch.

The quick reply / writing SMS isn't implemented in libwatchfish as far as I know. 

Answering or rejecting phone call can be done using:
https://github.com/piggz/harbour-amazfish/blob/cbe32738dc9dafa1a787bc41bf100a4037c70620/daemon/src/devices/pinetimejfdevice.cpp#L492

I had some issues with current interface, because it doesn't indicate current status of the phone. For example: I have rejected phone call on the phone, but smartwatch dialog wasn't dismissed, because information about status change isn't passed to smartwatch. I had some PoC implementation, but it was buggy so I have postponed it.

thmichel:
Answering and rejecting a phone call on the watch works without issues, which is most important to me. Also the watch stops showing an incoming call when it's answered or rejected on the phone, so that's all good.

Garmin has a feature to send a quick reply via SMS when rececting a call; I receive that request in Amazfish but wasn't able to trigger sendig an SMS via dbus. This isn't a big deal for me, probably something to look at in the future. It's quite handy when you're in a meeting or on the bike to tell someone you're busy / will call back. 

Overall I'm quite happy with where I am right now. Next things will be calendar sync (almost done) and weather requests. Once this works I'll look at downloading data to have historic data in Amazfish.

thmichel:
I noticed there were some errors compiling for Ubuntu touch, I tried to fix them.

PR #635, opened 2026-06-05
[Bangle.js] delete synced and 30days old sport activity from device

jmlich:
I would like to check it more carefully, but I have deleted all data from my device ;-)

### DIFF
--- a/daemon/src/devices/banglejsdevice.cpp
+++ b/daemon/src/devices/banglejsdevice.cpp
@@ -354,13 +354,10 @@ void BangleJSDevice::downloadSportsData() {
         return;
     }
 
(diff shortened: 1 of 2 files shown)


daemon/src/devices/banglejsdevice.cpp
```diff
@@ -354,13 +354,13 @@ void BangleJSDevice::downloadSportsData() {
         return;
     }
 
-    QString lastSyncId = AmazfishConfig::instance()->value("device/lastsportsyncid").toString();
+    // QString lastSyncId = AmazfishConfig::instance()->value("device/lastsportsyncid").toString();
```
piggz: may aswell remove these commented lines?

PR #636, opened 2026-06-05
Make supportsDataType() available as DaemonInterface function 

jmlich:
same as supportsFeature(), hide unsupported datagraphs.
Additionally, I have noticed redundant functions in harbour-amazfish.qml. Not sure if I should reduce number of functions in this case.

### DIFF
--- a/ui/qml/harbour-amazfish.qml
+++ b/ui/qml/harbour-amazfish.qml
@@ -216,16 +216,6 @@ ApplicationWindowPL
         popup.showMessage(msg)
     }
 
-    function supportsFeature(feature) {
-        console.log("Supports feature", feature, (_authenticated && !_authenticated) || ((DaemonInterfaceInstance.supportedFeatures() & feature) === feature));
-        return (_authenticated && !_authenticated) || ((DaemonInterfaceInstance.supportedFeatures() & feature) === feature)
-    }
-
-    function supportsData(data) {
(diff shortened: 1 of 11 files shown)


piggz:
I was actually thinking of removing the supportsFeature() function .. and just using the supported* versions? thoughts?

ui/qml/harbour-amazfish.qml
```diff
@@ -216,16 +216,6 @@ ApplicationWindowPL
         popup.showMessage(msg)
     }
 
-    function supportsFeature(feature) {
-        console.log("Supports feature", feature, (_authenticated && !_authenticated) || ((DaemonInterfaceInstance.supportedFeatures() & feature) === feature));
-        return (_authenticated && !_authenticated) || ((DaemonInterfaceInstance.supportedFeatures() & feature) === feature)
-    }
```
jmlich, line 222: I was looking into that function.
If I remember correctly the `(_authenticated && !_authenticated)` should provide way to refresh it when _authenticated is changed, but In the function it is evaluated just when function is triggered. The property binding is not in use.
I think this can be safely removed without change of behaviour.

jmlich:
Originally, I tried to add a supportsDataType() check to AnalysisPage, because graphs were also being displayed for devices that don't support the corresponding data type.

Regarding your question about the future of supportsFeature() and supportsDataType(), the changes I just proposed show the direction I would personally prefer.

PR #638, opened 2026-06-15
Move service parsing into the abstract device

piggz:
This removes the need for every device to implement the XML parsing, and removes a lot of duplication.

Devies now just need to implement a drv_createService function that will create a QBLEService if the device supports that UUID.

### DIFF
--- a/daemon/src/devices/abstractdevice.cpp
+++ b/daemon/src/devices/abstractdevice.cpp
@@ -100,8 +100,8 @@ void AbstractDevice::parseServices()
             QDBusInterface devInterface("org.bluez", path, "org.bluez.GattService1", QDBusConnection::systemBus(), 0);
             QString uuid = devInterface.property("UUID").toString();
 
-            qDebug() << "Creating service for: " << uuid;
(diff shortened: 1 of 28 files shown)


daemon/src/devices/abstractdevice.cpp
```diff
@@ -72,6 +73,45 @@ void AbstractDevice::devicePairFinished(const QString &status)
...
+
+        if (nodeName.startsWith("service")) {
+            QString path = devicePath() + "/" + nodeName;
+
+            QDBusInterface devInterface("org.bluez", path, "org.bluez.GattService1", QDBusConnection::systemBus(), 0);
+            QString uuid = devInterface.property("UUID").toString();
+
+            qDebug() << "Creating service for: " << uuid;
+            QBLEService *service = drv_createService(uuid, path);
+
```
jmlich: It would be great to move qDebug() one line down and add information if the service was resolved
```
            qDebug() << "Creating service for: " << uuid << service; 
```
or maybe
```
            qDebug() << "Creating service for: " << uuid << (service != nullptr); 
```

PR #639, opened 2026-06-21
Zeppos HRM and step goal

b100dian:
Except the [-1 smart](https://codeberg.org/Freeyourgadget/Gadgetbridge/src/branch/master/app/src/main/java/nodomain/freeyourgadget/gadgetbridge/service/devices/huami/zeppos/ZeppOsSupport.java#L370-L372) which I probably don't have and would require UI change.

Should save: HRM sleep detection, HRM interval and Step goal on the device.

### DIFF
--- a/daemon/src/devices/huami/zeppos/zepposconfigservice.cpp
+++ b/daemon/src/devices/huami/zeppos/zepposconfigservice.cpp
@@ -1,5 +1,6 @@
 #include "zepposconfigservice.h"
 #include "amazfishconfig.h"
+#include "typeconversion.h"
 
 ZeppOsConfigService::ZeppOsConfigService(ZeppOSDevice *device) : AbstractZeppOsService(device, false)
 {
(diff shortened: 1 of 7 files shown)


daemon/src/devices/huami/zeppos/zepposconfigservice.cpp
```diff
@@ -0,0 +1,174 @@
...
+        uint8_t configGroup = payload[3 + i];
+        qDebug() << "Got supported config group:" << QString::number(configGroup, 16);
+        requestConfig(configGroup);
+    }
+}
+
+void ZeppOsConfigService::requestConfig(uint8_t group)
+{
+    QByteArray cmd;
+    cmd += char(CMD_REQUEST);
```
piggz, line 73: The convention in the rest of the code is to use the macros like
UCHAR_TO_BYTEARRAY
to do such conversions.

Is there a benefit to just casting to char like this i wonder?

b100dian: I tried but it crashes, probably because it'd take the address of a const (switched to constexpr now). I did change other conversions to TypeConversion.
`char` is the type accepted by the `+=` operator.
Want me to change the cast to `(char)` for example?

ui/qml/pages/Settings-user.qml
```diff
@@ -200,8 +200,8 @@ PagePL {
 
             minimumValue: 0
```
b100dian, line 201: set -5 here to test 'smart' on newer ZeppOS

PR #640, opened 2026-06-23
Replace the download* functions with the fetchData interface

piggz:


ui/qml/pages/DebugInfo.qml
```diff
@@ -297,14 +297,6 @@ PagePL {
                 immediateAlertLevel = (immediateAlertLevel + 1) % 2;
             }
         }
-        ButtonPL {
-            text: qsTr("Fetch debug log")
-            anchors.horizontalCenter: parent.horizontalCenter
-            width: parent.width * 0.8
-            onClicked: {
-                DaemonInterfaceInstance.fetchLogs();
```
jmlich, line 305: Why is this removed? I would expect this
 DaemonInterfaceInstance.fetchData(Amazfish.TYPE_DEBUGLOG)

piggz: I thought about that, but there is a function below for general fetching of anything, which can fetch the debug logs also

PR #643, opened 2026-07-11
Fix authentication on Bip

vige:
Service creation refactoring broke authentication on original Bip, this commit fixes it

### DIFF
--- a/daemon/src/devices/huami/bipdevice.h
+++ b/daemon/src/devices/huami/bipdevice.h
@@ -95,6 +95,7 @@ public:
 
 protected:
     void initialise() override;
+    QBLEService* drv_createService(const QString &uuid, const QString &path) override;
 
 };
 
--- a/daemon/src/devices/huami/bipdevice.cpp
+++ b/daemon/src/devices/huami/bipdevice.cpp
@@ -91,6 +91,17 @@ void BipDevice::initialise()
     }
 }
 
+QBLEService *BipDevice::drv_createService(const QString &uuid, const QString &path)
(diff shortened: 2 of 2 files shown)


piggz:
Not sure this is correcct ... isnt HuamiDevice::drv_createService enough?

vige:
> Not sure this is correcct ... isnt HuamiDevice::drv_createService enough?

The parameters for MiBand2Service constructor are wrong for Bip in HuamiDevice.

piggz:
Ah ok, I understand ... in that case, can you just handle the one service in the Bip, and call the huamidevice for the rest if not MiBandService2?

vige:
> Ah ok, I understand ... in that case, can you just handle the one service in the Bip, and call the huamidevice for the rest if not MiBandService2?

Sure, that just requires a bit more refactoring. I can do that maybe tomorrow.

vige:
Didn't really need any refactoring, I don't know what I was thinking yesterday

piggz:
Yeah, this is kinda how i expected it to look :)

daemon/src/devices/huami/bipdevice.h
```diff
@@ -95,6 +95,7 @@ class BipDevice : public HuamiDevice
 
 protected:
     void initialise() override;
+    QBLEService* drv_createService(const QString &uuid, const QString &path);
```
piggz: Missing override here?

vige: added

PR #647, opened 2026-07-21
ZeppOS Sleepsession

piggz:
Adds support for fetching and displaying sleep sessions from ZeppOS devices.

Data is saved as raw base64 and decoded at runtime, this would allow other devices to re-use the table.

Currently not all features of the sleep-sessions are displayed in the UI, just basic hours.  Its possible to pull out each period type and display on a timeline.

DataSource is refactored to allow other device types to provide their own implementation.

TypeConverion moved to lib/ to allow the UI to use.

### DIFF
--- a/daemon/CMakeLists.txt
+++ b/daemon/CMakeLists.txt
@@ -242,6 +242,7 @@ set(HEADERS
 )
 
 add_executable(harbour-amazfishd ${SOURCES} ${HEADERS}
+    src/operations/huami/fetchsleepsessionoperation.h src/operations/huami/fetchsleepsessionoperation.cpp
 )
 
 find_package(PkgConfig REQUIRED)
--- a/daemon/CMakeLists.txt
+++ b/daemon/CMakeLists.txt
@@ -115,7 +115,6 @@ set(SOURCES
 ./src/services/pinetimesimpleweatherservice.cpp
 ./src/services/uartservice.cpp
 ./src/transliterator.cpp
-./src/typeconversion.cpp
 ./src/ecdh/ecdh.c
 )
 
@@ -237,7 +236,6 @@ set(HEADERS
 ./src/services/pinetimesimpleweatherservice.h
 ./src/services/uartservice.h
 ./src/transliterator.h
-./src/typeconversion.h
 ./src/ecdh/ecdh.h
 )
 
--- a/daemon/src/devices/huami/huamifetcher.cpp
+++ b/daemon/src/devices/huami/huamifetcher.cpp
@@ -54,7 +54,7 @@ void HuamiFetcher::startFetchData(Amazfish::DataTypes type)
     if (type & Amazfish::DataType::TYPE_TEMPERATURE && m_device->supportsDataType(Amazfish::DataType::TYPE_TEMPERATURE)) {
         m_operations.append(new FetchTemperatureOperation(this, m_device->database(), m_device->isZeppOs()));
     }
-    if (type & Amazfish::DataType::TYPE_SLEEP && m_device->supportsDataType(Amazfish::DataType::TYPE_SLEEP)) {
+    if (type & Amazfish::DataType::TYPE_SLEEP && m_device->supportsDataType(Amazfish::DataType::TYPE_SLEEP) && m_device->isZeppOs()) {
         m_operations.append(new FetchSleepSessionOperation(this, m_device->database()));
     }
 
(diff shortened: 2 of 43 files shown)


daemon/src/devices/huami/huamifetcher.cpp
```diff
@@ -53,6 +54,9 @@ void HuamiFetcher::startFetchData(Amazfish::DataTypes type)
     if (type & Amazfish::DataType::TYPE_TEMPERATURE && m_device->supportsDataType(Amazfish::DataType::TYPE_TEMPERATURE)) {
         m_operations.append(new FetchTemperatureOperation(this, m_device->database(), m_device->isZeppOs()));
     }
+    if (type & Amazfish::DataType::TYPE_SLEEP && m_device->supportsDataType(Amazfish::DataType::TYPE_SLEEP) && m_device->isZeppOs()) {
```
jmlich: this condition feels somehow redundant, but I don't know background of this decision right now. It looks like that supportsDataType(TYPE_SLEEP) should be enough

daemon/CMakeLists.txt
```diff
@@ -237,11 +236,11 @@ set(HEADERS
 ./src/services/pinetimesimpleweatherservice.h
 ./src/services/uartservice.h
 ./src/transliterator.h
-./src/typeconversion.h
 ./src/ecdh/ecdh.h
 )
 
 add_executable(harbour-amazfishd ${SOURCES} ${HEADERS}
+    src/operations/huami/fetchsleepsessionoperation.h src/operations/huami/fetchsleepsessionoperation.cpp
```
jmlich, line 243: Why are those directly added to add_executable, why not SOURCES and HEADERS?

ui/src/datasource.h
```diff
@@ -34,17 +34,24 @@ class DataSource: public QObject
...
 
-private:
-    KDbConnection *m_conn = nullptr;
-    
     struct SleepSession {
         QDateTime sleepStart;
         QDateTime sleepEnd;
         long lightSleepDuration;
         long deepSleepDuration;
+        long remSleepDuration;
```
jmlich, line 42: claude suggests some default initialization
        long remSleepDuration = 0

ui/src/dataproviderfactory.cpp
```diff
@@ -0,0 +1,28 @@
...
+using DataCreator = std::function<DataSource*()>;
+
+static const QMap<QString, DataCreator> dataMap = {
+    { "amazfitbip", []() { return new HuamiDataSource(); } },
+    { "amazfitbiplite", []() { return new HuamiDataSource(); } },
+    { "amazfitgtr", []() { return new HuamiDataSource(); } },
+    { "amazfitgtr2", []() { return new HuamiDataSource(); } },
+    { "amazfitgts", []() { return new HuamiDataSource(); } },
+    { "amazfitgts2", []() { return new HuamiDataSource(); } },
+    { "amazfitneo", []() { return new HuamiDataSource(); } },
```
jmlich, line 14: I think I was using HuamiDataSource::calculateSleep() for banglejs.

jmlich (commented):
In general, I find this approach to obtaining the data source a bit difficult to understand. I’m not against the change itself, but I think it would be great if the daemon could store the data in the same structure for every device and handle the necessary data conversions as well. Ideally, the UI should only need to visualize the data rather than handle device-specific conversions. I think these points could be considered as a future improvement.

ui/src/dataproviderfactory.cpp
piggz: I could implement the sleep session data as a separate child table, which I agree would certainly be neater, i was being lazy and copying the GB method!  The implementation would have to be based around zeppos, though I think that would be mostly ok to generically represent sleep.

jmlich: I mean, separate child table isn't necessary. We should just verify, that everything works as before.
In my case `{ "banglejs", []() { return new HuamiDataSource(); } },` should be probably added into that factory. I may need to check if it still works.

piggz: the extra table wasnt in the context of the DataSource() .. it was in the context of storing sleep data.  currently, there is a sleep record each day, and in theat record is a blob of data.  that blob of data contains multiple 'sleep sessions' which describe the sleep type, time and duration.  So, to not have the blob record would require another table containing multiple sleep-sessions per day

PR #656, opened 2026-09-05
Store device as only the address and find it on any hci device

### LINKED ISSUE #634: No bluetooth connection possible when using Bloutooth Support for AAS in SailfishOS 5.1
When enabling bluetooth support in SailfishOs 5.1, Amazfish cannot connect to any device any more. Error message is "Method Connect with Signature "" doesn't exist"

It seems a second adapter is added in this configuration, the application settings page shows  /org/bluez/hci1" as adapter, while the device is discovered on hci0.

I scanned through the code and could not find a reference to the bluetooth adapter except in the "unpair()" function of the qbledevice, where it uses the static configuration "/org/bluez/hci0". 

I couldn't find the part of the code where it changes the adapter to hci1.

piggz:
Fixes: #634
Stores the device address as the raw BT mac accress, then searches for it on any hci device.  Removes the hci device selection from the UI as shouldnt be required.

### DIFF
--- a/daemon/src/deviceinterface.cpp
+++ b/daemon/src/deviceinterface.cpp
@@ -153,13 +153,12 @@ void DeviceInterface::disconnect()
 void DeviceInterface::unpair()
 {
     qDebug() << Q_FUNC_INFO;
-    if (m_device) {
+    if (m_device && !m_adapterPath.isEmpty()) {
         BluezAdapter adapter;
-        adapter.setAdapterPath(AmazfishConfig::instance()->localAdapter());
+        adapter.setAdapterPath(m_adapterPath);
         adapter.removeDevice(m_deviceAddress);
         delete m_device;
         m_device = nullptr;
-
     }
 }
 
@@ -739,6 +738,21 @@ void DeviceInterface::log_battery_level(int level) {
 QString DeviceInterface::devicePath(const QString &address)
 {
     qDebug() << Q_FUNC_INFO << address;
+
+    if (!determinAdapterPath(address)){
+        qDebug() << "No device path found";
+        return QString();
+    }
+
+    QString formattedAddress = address;
+    formattedAddress.replace(":", "_");
+
+    return m_adapterPath + "/dev_" + formattedAddress;
+}
+
+bool DeviceInterface::determinAdapterPath(const QString &address)
+{
+    qDebug() << Q_FUNC_INFO << address;
     AdapterModel adapterModel;
 
     for (int i = 0; i < adapterModel.rowCount(); ++i) {
@@ -749,16 +763,18 @@ QString DeviceInterface::devicePath(const QString &address)
         QString deviceString = adapter["itemText"].toString() + "/dev_" + formattedAddress;
         qDebug() << adapter << deviceString;
 
-        BluezAdapter *bluezAdapter = new BluezAdapter();
-        bluezAdapter->setAdapterPath(adapter["itemText"].toString());
+        BluezAdapter bluezAdapter;
+        bluezAdapter.setAdapterPath(adapter["itemText"].toString());
 
-        if (bluezAdapter->deviceIsValid(deviceString)) {
-            return deviceString;
+        if (bluezAdapter.deviceIsValid(deviceString)) {
+            m_adapterPath = adapter["itemText"].toString();
+            AmazfishConfig::instance()->setLocalAdapter(m_adapterPath); //Used by PTJF Device for local server
+            return true;
         }
     }
 
     qDebug() << "No device path found";
-    return QString();
+    return false;
 }
 
 void DeviceInterface::slot_informationChanged(Amazfish::Info key, const QString &val)
--- a/daemon/src/deviceinterface.cpp
+++ b/daemon/src/deviceinterface.cpp
@@ -8,6 +8,7 @@
 #include <QDir>
 #include <QFile>
 #include <QProcess>
+#include <adaptermodel.h>
 
 #include <KDb3/KDbDriverManager>
 #include <KDb3/KDbTransactionGuard>
@@ -100,9 +101,10 @@ void DeviceInterface::connectToDevice(const QString &address)
 {
     qDebug() << Q_FUNC_INFO << ": address:" << address;
 
-    if (m_device) {
-        m_deviceAddress = address;
-        m_device->setDevicePath(address);
+    m_deviceAddress = devicePath(address);
+
+    if (m_device && !m_deviceAddress.isEmpty()) {
+        m_device->setDevicePath(m_deviceAddress);
         m_device->connectToDevice();
     }
     else {
@@ -115,7 +117,7 @@ QString DeviceInterface::pair(const QString &name, const QString &deviceType, co
 {
     qDebug() << Q_FUNC_INFO << name << deviceType << address;
 
-    m_deviceAddress = address;
+    m_deviceAddress = devicePath(address);
 
     if (m_device) {
         delete m_device;
@@ -123,7 +125,7 @@ QString DeviceInterface::pair(const QString &name, const QString &deviceType, co
     m_device = DeviceFactory::createDevice(name, deviceType);
 
     if (m_device) {
-        m_device->setDevicePath(address);
+        m_device->setDevicePath(m_deviceAddress);
         connect(m_device, &AbstractDevice::connectionStateChanged, this, &DeviceInterface::onConnectionStateChanged, Qt::UniqueConnection);
         connect(m_device, &AbstractDevice::message, this, &DeviceInterface::message, Qt::UniqueConnection);
         connect(m_device, &AbstractDevice::downloadProgress, this, &DeviceInterface::downloadProgress, Qt::UniqueConnection);
@@ -732,6 +734,31 @@ void DeviceInterface::log_battery_level(int level) {
     }
     tg.commit();
 
+}
+
+QString DeviceInterface::devicePath(const QString &address)
+{
+    qDebug() << Q_FUNC_INFO << address;
+    AdapterModel adapterModel;
+
+    for (int i = 0; i < adapterModel.rowCount(); ++i) {
+        QVariantMap adapter = adapterModel.get(i);
+        QString formattedAddress = address;
+        formattedAddress.replace(":", "_");
+
+        QString deviceString = adapter["itemText"].toString() + "/dev_" + formattedAddress;
+        qDebug() << adapter << deviceString;
+
+        BluezAdapter *bluezAdapter = new BluezAdapter();
+        bluezAdapter->setAdapterPath(adapter["itemText"].toString());
+
+        if (bluezAdapter->deviceIsValid(deviceString)) {
+            return deviceString;
+        }
+    }
+
+    qDebug() << "No device path found";
+    return QString();
 }
 
 void DeviceInterface::slot_informationChanged(Amazfish::Info key, const QString &val)
--- a/daemon/src/deviceinterface.cpp
+++ b/daemon/src/deviceinterface.cpp
@@ -87,6 +87,13 @@ DeviceInterface::DeviceInterface()
 
     //Finally, connect to device if it is defined
     QString pairedAddress = config->pairedAddress();
+
+    //Convert old format address to new
+    if (pairedAddress.contains("/org/bluez/hci")) {
+        pairedAddress = pairedAddress.right(17).replace("_", ":");
+        config->setPairedAddress(pairedAddress);
+    }
+
     if (!pairedAddress.isEmpty()) {
         connectToDevice(pairedAddress);
     }
--- a/daemon/src/deviceinterface.h
+++ b/daemon/src/deviceinterface.h
@@ -114,6 +114,8 @@ private:
 
     void log_battery_level(int level);
 
+    QString devicePath(const QString &address);
+
     HRMService *hrmService() const;
     
     Q_SLOT void onNotification(watchfish::Notification *notification);
--- a/daemon/src/deviceinterface.h
+++ b/daemon/src/deviceinterface.h
@@ -94,6 +94,8 @@ private:
     int m_connectionStateChangedCount = 0;
     QString m_deviceAddress;
     QString m_deviceName;
+    QString m_adapterPath;
+
     bool m_dbusRegistered = false;
     int m_lastBatteryLevel = 0;
     int m_lastAlertHash = 0;
@@ -115,6 +117,7 @@ private:
     void log_battery_level(int level);
 
     QString devicePath(const QString &address);
+    bool determinAdapterPath(const QString &address);
 
     HRMService *hrmService() const;
     
--- a/ui/qml/pages/PairDevicePage.qml
+++ b/ui/qml/pages/PairDevicePage.qml
@@ -139,7 +139,7 @@ PageListPL {
 
                 _deviceName = model.FriendlyName;
                 _deviceType = (enforcedDeviceType !== undefined) ? enforcedDeviceType.deviceType : device.deviceType
-                _deviceAddress = AmazfishConfig.localAdapter+"/dev_" + model.Address.replace(/:/g, '_');
+                _deviceAddress = model.Address;
 
                 if (device_auth) {
                     var authdialog = app.pages.push(Qt.resolvedUrl("./AuthKeyDialog.qml"));
--- a/ui/qml/pages/Settings-app.qml
+++ b/ui/qml/pages/Settings-app.qml
@@ -17,70 +17,6 @@ PagePL {
         anchors.top: parent.top
         anchors.margins: styler.themePaddingMedium
 
-        AdapterModel {
-            id: adapters
-        }
-
-        ComboBoxPL {
-            id: cboLocalAdapter
-            model: adapters
-            textRole: "path"
-            label: qsTr("BT Adapter")
-            Component.onCompleted: {
-                for (var i = 0; i < adapters.rowCount(); i++) {
-                    var item = adapters.get(i)
-                    if (item.path === AmazfishConfig.localAdapter) {
-                        cboLocalAdapter.currentIndex = i;
-                        setDetails(item)
-                        return
-                    }
-                }
-            }
-
-            onCurrentIndexChanged: {
-                setDetails(adapters.get(currentIndex))
-            }
-
-            function setDetails(data) {
-                lblBluetoothName.text = data.name
-                lblBluetoothAlias.text = data.alias
-                lblBluetoothAddress.text = data.address
-            }
-        }
-
-
-        LabelPL {
-            id: lblBluetoothName
-            color: styler.themeSecondaryHighlightColor
-            font.pixelSize: styler.themeFontSizeMedium
-            truncMode: truncModes.fade
-            anchors.left: parent.left
-            anchors.right: parent.right;
-            anchors.leftMargin: styler.themePaddingMedium
-            anchors.rightMargin: styler.themePaddingMedium
-        }
-
-        LabelPL {
-            id: lblBluetoothAlias
-            color: styler.themeSecondaryHighlightColor
-            font.pixelSize: styler.themeFontSizeMedium
-            truncMode: truncModes.fade
-            anchors.left: parent.left
-            anchors.right: parent.right;
-            anchors.leftMargin: styler.themePaddingMedium
-            anchors.rightMargin: styler.themePaddingMedium
-        }
-        LabelPL {
-            id: lblBluetoothAddress
-            color: styler.themeSecondaryHighlightColor
-            font.pixelSize: styler.themeFontSizeMedium
-            truncMode: truncModes.fade
-            anchors.left: parent.left
-            anchors.right: parent.right;
-            anchors.leftMargin: styler.themePaddingMedium
-            anchors.rightMargin: styler.themePaddingMedium
-        }
-
         SectionHeaderPL {
             visible: supportsFeatureRefresh(Amazfish.FEATURE_ALERT)
             text: qsTr("Notifications")
@@ -267,7 +203,6 @@ PagePL {
         AmazfishConfig.appNotifyLowBattery = chkNotifyLowBattery.checked;
         AmazfishConfig.appNavigationNotification = chkNavigationNotification.checked;
         AmazfishConfig.appSimulateEventSupport = chkSimulateEventSupport.checked;
-        AmazfishConfig.localAdapter = cboLocalAdapter.value;
         AmazfishConfig.appTransliterate = chkTransliterate.checked ;
     }
 


piggz:
This should fix connection on sailfish devices with app-support BT bridge, which causes the hci device id to change.

jmlich:
I have asked claude for review. Its response looks valid.

  1. Startup race: a failed adapter lookup kills the daemon's connection permanently
 
  daemon/src/deviceinterface.cpp:107

  m_deviceAddress = devicePath(address);
  if (m_device && !m_deviceAddress.isEmpty()) {
      m_device->setDevicePath(m_deviceAddress);
      m_device->connectToDevice();
  }

  devicePath() returns empty whenever BlueZ doesn't yet expose the device — adapter still powering up, bluetoothd started after the daemon, watch out of range at boot. In that case AbstractDevice::connectToDevice() is never entered, and
  that is the only place m_reconnectTimer->start() is called (daemon/src/devices/abstractdevice.cpp:35-45). So the 60s auto-reconnect never arms and the daemon sits dead until the user manually taps "Connect to watch".

  Before this change the path was built blindly and the connect was always attempted, so BlueZ errors were transient and the retry loop covered them. Suggestion: on an empty path, still start the retry loop (or retry devicePath() on a
  timer) instead of bailing out. The user-facing text is also wrong for this case — "Device is not valid, it may not be supported" for what is really "adapter/device not available yet".

  2. device_id in the database changes format — existing history is orphaned

  qHash(config->pairedAddress()) is the device_id column:

  - daemon/src/devices/banglejsdevice.cpp:224,754,799
  - daemon/src/devices/pinetimejfdevice.cpp:199
  - daemon/src/operations/huami/activityfetchoperation.cpp:66
  - daemon/src/operations/huami/sportsdetailoperation.cpp:112

  Rewriting pairedAddress from /org/bluez/hci0/dev_XX_… to XX:XX:… (deviceinterface.cpp:92-95) changes that hash, so every existing mi_band_activity row becomes invisible — today's step count resets, charts go blank, and re-synced data
  lands under a new id. The migration block is the right place to also do:

  UPDATE mi_band_activity SET device_id = <new> WHERE device_id = <old>;

  (compute both hashes before overwriting the config value). Same for any other table keyed on device_id.

  3. pair() doesn't apply the same guard as connectToDevice()

  daemon/src/deviceinterface.cpp:124-137 — if devicePath() returns empty, the code still creates the device, calls setDevicePath(""), m_device->pair(), and returns "pairing". The UI then waits forever on a state change that can't come.
  Mirror the emptiness check and return the error string.

  4. Smaller things

  - determinAdapterPath → determineAdapterPath (deviceinterface.h:120, .cpp:756). Typo in new API.
  - formattedAddress is recomputed inside the loop (.cpp:764-766) and again in devicePath() (.cpp:750-753). Hoist it once and pass it in; deviceString is also built before the adapter it belongs to is set.
  - Per-call D-Bus leaks. Each determinAdapterPath() constructs an AdapterModel (leaks its new QDBusInterface — no parent, default dtor) plus one BluezAdapter::setAdapterPath() per adapter (same leak). Pre-existing in those classes, but
    this change turns a one-off into a per-connect leak. Giving those QDBusInterfaces a parent fixes it at the source.
  - #include <adaptermodel.h> (.cpp:11) — angle brackets, and grouped with the Qt includes. bluezadapter.h right above uses quotes; match it.
  - The migration only runs in the daemon. ui/qml/pages/FirstPage.qml:49 passes AmazfishConfig.pairedAddress straight to connectToDevice. If the UI reads the config before the daemon has rewritten it, it hands back an old-style path,
    which devicePath() mangles into …/hci0/dev_/org/bluez/hci0/dev_XX… and silently fails. Cheap defence: have devicePath() normalise an input that starts with /org/bluez instead of assuming a bare address.
  - qmlRegisterType<AdapterModel> (ui/src/harbour-amazfish-ui.cpp:108) is now unused by QML since the settings combo is gone — harmless, but droppable if nothing else needs it.

jmlich:
1. determineAdapterPath() is now passed the address with colons — nothing ever matches

  daemon/src/deviceinterface.cpp:783-808
```
  QString DeviceInterface::devicePath(const QString &address)
  {
      QString formattedAddress = address;
      formattedAddress.replace(":", "_");
   
      if (!determineAdapterPath(address)){        // <-- raw address, not formattedAddress
          ...
      }   
      return m_adapterPath + "/dev_" + formattedAddress;
  }

  bool DeviceInterface::determineAdapterPath(const QString &address)
  {
      ...
      QString deviceString = adapter["itemText"].toString() + "/dev_" + address;   // dev_AA:BB:CC:...
      if (bluezAdapter.deviceIsValid(deviceString)) {
```   
  Hoisting formattedAddress out of the loop dropped the substitution from the path that's actually compared. BluezAdapter::deviceIsValid() matches against real BlueZ node names (dev_AA_BB_CC_DD_EE_FF), so dev_AA:BB:… never matches,
  devicePath() always returns empty, and both connectToDevice() and pair() always fail with "Device is not yet available". Pass formattedAddress into determineAdapterPath() (and drop the now-redundant replace there).

  2. After the migration, the old address is still used for the connect

  daemon/src/deviceinterface.cpp:117-138
```  
  QString pairedAddress = config->pairedAddress();
  if (pairedAddress.contains("/org/bluez/hci")) {
      QString newAddress = pairedAddress.right(17).replace("_", ":");
      migrateDataDeviceAddress(pairedAddress, newAddress);
      config->setPairedAddress(newAddress);      // config updated…
  }   
  if (!pairedAddress.isEmpty()) {
      connectToDevice(pairedAddress);            // …but the local is still the old path
  }   
```
  The previous version reassigned pairedAddress. Now the first connect after upgrade is handed /org/bluez/hci0/dev_XX_…, which devicePath() mangles into …/hci0/dev_/org/bluez/hci0/dev_XX…. It self-heals on the next 60s tick since the
  config was written, but the first attempt always fails. Add pairedAddress = newAddress;.
 
3. pair() is missing the return — commit 8f9bcca3 has no effect

  daemon/src/deviceinterface.cpp:155-159
```
  if (m_deviceAddress.isEmpty()) {
      qDebug() << "Device is not available";
      message(tr("Device is not yet available"));
  }   
  // falls through: creates the device, setDevicePath(""), m_device->pair(), returns "pairing"
```
  Needs return QString("no device found"); (or similar) so the UI isn't left waiting on a state change that can't arrive.

  4. migrateDataDeviceAddress() dereferences m_conn unguarded

  daemon/src/deviceinterface.cpp:1184-1190

  setupDatabase() returns early — leaving m_conn == nullptr — when no KDb driver is found, the driver fails to load, or connect() fails (deviceinterface.cpp:640-676). Every other DB user in the daemon guards with if (!m_conn || 
  !m_conn->isDatabaseUsed()). This one is reached from the constructor via reconnectionTimer() → connectToDevice(), so it's a startup crash on any install where the sqlite driver is missing. Add the same guard — and note that skipping
  the migration while still writing the new pairedAddress would silently orphan the data, so it's worth bailing out of the rewrite too rather than just the UPDATEs. 
 
  5. The reconnect timer can abort an in-progress pairing and connect to the previous watch

  daemon/src/deviceinterface.cpp:141-150
```
  if ((connectionState() != "authenticated" && m_autoreconnect) || connectionState() == "authfailed") {
      disconnect();
      connectToDevice();
  }   
 ```
  pair() sets m_autoreconnect = true and the state goes to "pairing". Pairing on AsteroidOS/Pebble/PineTime needs a confirmation on the watch, which can easily exceed 60s. When the timer fires it tears the session down and calls the
  no-arg connectToDevice(), which reads config->pairedAddress() — still the previously paired device, because PairDevicePage.qml:67 only writes it on "paired"/"connected". Exclude the in-progress states ("pairing", "connecting",
  "connected", "paired") from the retry condition, or suspend the timer while pair() is outstanding.

  Same condition, smaller: connectionState() returns QString() when m_device is null, so a fresh install with nothing paired runs the disconnect/connect dance every minute for no reason.
  
  6. The 60s retry makes the adapter scan's cost and leaks permanent

  Each failed tick runs determineAdapterPath(), which constructs an AdapterModel (blocking GetManagedObjects) plus one BluezAdapter per adapter (blocking Introspect each). Both classes new a QDBusInterface with no parent and never delete
  it (qble/adaptermodel.cpp:16, qble/bluezadapter.cpp:16), so a watch that's out of range overnight leaks a few hundred QDBusInterface objects and does two blocking system-bus round trips a minute. Giving those interfaces a parent fixes
  the leak at the source.
 
  Also user-visible: connectToDevice(address) emits message(tr("Device is not yet available")) on every failure, so the UI now gets a toast every 60 seconds while the watch is away. Worth suppressing on the timer-driven path.
7. Overloaded connectToDevice on the D-Bus object
  
  registerDBus() uses ExportAllInvokables (deviceinterface.cpp:1323), and both connectToDevice() and connectToDevice(QString) are Q_INVOKABLE. QtDBus dispatches by signature so this most likely works, but overloads on an exported object
  are a known rough edge and worth testing explicitly. Since FirstPage.qml now calls the no-arg form and PairDevicePage.qml goes through pair(), nothing calls the one-arg version over D-Bus any more — making it plain public instead of
  Q_INVOKABLE removes the question.
  
  Nits
 
  - oldAdress → oldAddress (deviceinterface.h:185, .cpp:1184).
  - qDebug() << Q_FUNC_INFO;; — stray semicolon (.cpp:119).
  - SET device_id='%1' quotes a value into an Integer/Unsigned column (.cpp:310, :409). SQLite's type affinity saves it, but dropping the quotes matches how the column is written everywhere else.
  - unpair() doesn't clear m_autoreconnect or m_adapterPath. Harmless today because FirstPage.qml:13 calls disconnect() first, but it leaves the invariant resting on caller order.
 
  Cleanly resolved from the last round: the include style, the determin→determine rename, the sports_data/mi_band_activity migration coverage, and the m_reconnectTimer removal across all six device classes (no dangling references left).

piggz:
Added fixes for the highlighted issues, except the leak which will have to be in qble

jmlich:
Fixed and verified: determineAdapterPath now gets the underscore form (d9becb13) — that was the blocker; the first connect after migration uses the new address (77ba85b5); pair() returns early (41b69274); the D-Bus overload is gone and
  the one-arg connectToDevice is private (eba6467c); the dead AdapterModel QML registration is removed (04dc7b6e). The master merge resolved the Bangle.js init guard cleanly — no duplicated block.

  Two things are still open, one of them newly introduced.

  1. Auto-reconnect is never re-armed after a fresh pair — new regression from 0d4610af

  m_autoreconnect is now written in exactly two places: set true in connectToDevice() when pairedAddress is non-empty (deviceinterface.cpp:135), set false in disconnect() (:190). 0d4610af removed the m_autoreconnect = true that pair()
  used to do.
 
  On a daemon that starts with nothing paired:
 
  1. Constructor calls reconnectionTimer(). connectionState() is QString() (no device), m_autoreconnect is true by default → the branch is taken.
  2. disconnect() sets m_autoreconnect = false.
  3. connectToDevice() finds pairedAddress empty → never sets it back.

  The user then pairs. pair() no longer touches the flag, and nothing else does — PairDevicePage.qml:67 only writes the config. So for the rest of that daemon's life m_autoreconnect stays false: the watch drops out of range and never
  reconnects until the user taps "Connect to watch" or restarts the daemon. It looks fine on the next daemon start, which is what makes it easy to miss in testing.

  Restoring m_autoreconnect = true after a successful m_device->pair() fixes it, but that reopens the reason it was removed — the timer firing mid-pairing. The clean fix is both: keep the flag set on pair, and make the timer skip states
  that are still making progress:

  const QString state = connectionState();
  if (state == "pairing" || state == "connecting" || state == "connected" || state == "paired") {
      return;
  }

  (As it stands the mid-pair teardown is actually already closed by accident — the only route to pair() with the flag set goes through FirstPage.qml:13's disconnect() — but that's an invariant resting on QML call order, not on the
  daemon.)

  2. migrateDataDeviceAddress() still dereferences m_conn unguarded
 
  deviceinterface.cpp:1184-1190 — unchanged from the last round. setupDatabase() returns early leaving m_conn == nullptr when no KDb driver is present, the sqlite driver fails to load, or connect() fails (:640-676), and the constructor
  reaches this via reconnectionTimer() → connectToDevice(). Every other DB caller in the daemon guards with if (!m_conn || !m_conn->isDatabaseUsed()).

  Worth guarding the rewrite, not just the two UPDATEs — if the migration can't run, writing the new pairedAddress anyway orphans the history permanently:

  if (!m_conn || !m_conn->isDatabaseUsed()) {
      qWarning() << Q_FUNC_INFO << "no database, deferring address migration";
      return false;   // caller skips setPairedAddress()
  }   

  3. Cost of the 60s retry while the watch is away (unchanged)

  Each failed tick runs determineAdapterPath(): one blocking GetManagedObjects plus a blocking Introspect per adapter, and both AdapterModel (qble/adaptermodel.cpp:16) and BluezAdapter (qble/bluezadapter.cpp:16) new a QDBusInterface with
  no parent and never delete it. A watch left out of range overnight leaks a few hundred of them. Parenting those two interfaces fixes it at the source.
  
  Also still user-visible: connectToDevice(address) emits message(tr("Device is not yet available")) on every failure (:112), so the UI gets a toast every minute while the watch is away. Suppressing it on the timer-driven path (a flag,
  or emitting only on the first failure of a run) is enough.

  Nits

  - return QString("device not available");; (:160) and qDebug() << Q_FUNC_INFO;; (:118) — stray semicolons.
  - oldAdress → oldAddress (deviceinterface.h:185, .cpp:1184).
  - 77ba85b5 re-reads config->pairedAddress() twice more after the migration block; assigning pairedAddress = newAddress inside the if and keeping the local read would be clearer and matches the rest of the function.
  - SET device_id='%1' quotes into an Integer/Unsigned column (:310, :409). SQLite's affinity handles it; unquoted matches how the column is written elsewhere.
  - unpair() (:196) still leaves m_autoreconnect and m_adapterPath untouched.
  - BangleJSDevice::m_initialised (banglejsdevice.h:111) duplicates AbstractDevice::m_hasInitialised (abstractdevice.h:106), which nothing currently uses. Not from this branch, but the two will drift.
  
  The QLoggingCategory / QFileInfo include commits are correct — -DUUITK_EDITION is defined as 1 (CMakeLists.txt:36), so the new #ifdef guard matches the existing #elif UUITK_EDITION at the use site.

jmlich:
Reviewed 7d33daa6..5f5076ab.
  
  Fixed and verified: auto-reconnect is re-armed after pairing and the in-progress states are excluded from the retry (5c981ef9) — I checked that "paired" is still reachable (huami/huamidevice.cpp:468), so that guard entry isn't dead,
  and that a failed BlueZ Connect emits QBLEDevice::error → "disconnected" (abstractdevice.cpp:105-110), so "connecting" can't wedge the timer. The m_conn guard is in (db4377d6), plus the oldAdress typo, both stray ;;, the quoted integer
  literals, and unpair() clearing m_autoreconnect/m_adapterPath (15b18b75). Removing the dead devicePairFinished is fine — its only reference was the commented-out connect.
  
  Two things in the new migration logic, then the carry-overs.

  1. migrateDataDeviceAddress() returns true even when the UPDATEs fail
 
  daemon/src/deviceinterface.cpp:1198-1212

  m_conn->executeSql(KDbEscapedString("UPDATE mi_band_activity SET device_id=%1 WHERE device_id=%2")...);
  m_conn->executeSql(KDbEscapedString("UPDATE sports_data      SET device_id=%1 WHERE device_id=%2")...);
  return true;
 
  Both return values are discarded, so the function really reports "a database connection exists", not "the data was migrated". If either UPDATE fails — or the first succeeds and the second doesn't — the caller still calls
  setPairedAddress(newAddress) and the un-migrated rows are orphaned permanently, which is exactly what db4377d6 was meant to prevent. Since the whole point is the all-or-nothing coupling with the config write, this wants both results
  and a transaction around them:
 
  KDbTransaction t = m_conn->beginTransaction();
  KDbTransactionGuard tg(t);
  bool ok = m_conn->executeSql(...);
  ok = m_conn->executeSql(...) && ok;
  if (ok) {
      tg.commit();
  }
  return ok;
  
  (KDbTransactionGuard is already included and used in createTables().)
 
  2. A failed migration now means the watch never connects at all                                                                                                                                                                            
                                                                                                                                                                                                                                             
  deviceinterface.cpp:127-136 — when migration returns false, pairedAddress is cleared and the connect is skipped. On a system where the KDb sqlite driver is missing or the DB is unwritable, that turns "watch works, history doesn't" into
  "watch doesn't work", every 60 seconds, forever, with no way out from the UI.                                                                                                                                                              
                                                                                                                                                                                                                                             
  It doesn't have to be that way: device_id is computed from config->pairedAddress() at write time, so connecting with the converted address held only in memory — without persisting it — keeps new samples landing on qHash(oldAddress),   
  consistent with the existing rows. Nothing gets split, and the migration retries on the next start:                                                                                                                                        
                                                                                                                                                                                                                                             
  QString newAddress = pairedAddress.right(17).replace("_", ":");                                                                                                                                                                            
  if (migrateDataDeviceAddress(pairedAddress, newAddress)) {                                                                                                                                                                                 
      config->setPairedAddress(newAddress);                                                                                                                                                                                                  
  }                                                                                                                                                                                                                                          
  pairedAddress = newAddress;   // connect either way                                                                                                                                                                                        
                                                                                                                                                                                                                                             
  Your call on the trade-off, but refusing to connect is the harsher of the two failure modes.                                                                                                                                               
                                                                                                                                                                                                                                             
  3. The guard uses the local, the call still reads the config                                                                                                                                                                               
                                                                                                                                                                                                                                             
  if (!pairedAddress.isEmpty()) {                                                                                                                                                                                                            
      m_autoreconnect = true;                                                                                                                                                                                                                
      connectToDevice(config->pairedAddress());   // <-- should be pairedAddress                                                                                                                                                             
  }                                                                                                                                                                                                                                          
                                                                                                                                                                                                                                             
  15b18b75 introduced the local pairedAddress = newAddress but left the call reading from config. They agree today in both branches, so there's no live bug — but the guard and the argument now come from different sources, which is the   
  kind of thing that breaks on the next edit. Use pairedAddress.                                                                                                                                                                             
                                                                                                                                                                                                                                             
  4. Carry-overs, unchanged                                                                                                                                                                                                                  
                                                                                                                                                                                                                                             
  Each failed tick still constructs an AdapterModel (blocking GetManagedObjects) and one BluezAdapter per adapter (blocking Introspect), and both new a QDBusInterface with no parent that is never deleted (qble/adaptermodel.cpp:16,       
  qble/bluezadapter.cpp:16) — a watch out of range overnight leaks a few hundred. And connectToDevice(address) still emits message(tr("Device is not yet available")) on every failure (:111), so the UI gets a toast every minute while the 
  watch is away.                                                                                                                                                                                                                             
                                                                                                                                                                                                                                             
  Nits                                                                                                                                                                                                                                       
                                                                                                                                                                                                                                             
  - 5f5076ab flips const QString &x → const QString& x in exactly two places (devicePath, migrateDataDeviceAddress), against 18 occurrences of the existing style in the same two files — and devicePath's header declaration still says     
    const QString &address while its definition now says const QString&. Given the commit is titled "Fix formatting errors", it's probably worth reverting those two rather than leaving the file mixed. Same commit also restyles //X → // X
    on lines it doesn't otherwise change.                                                                                                                                                                                                    
  - reconnectionTimer() caches const QString state and then calls connectionState() three more times in the next condition — reuse state.                                                                                                    
  - 5c981ef9 un-commented qDebug() << Q_FUNC_INFO; at the top of reconnectionTimer(), so it now logs unconditionally every 60 seconds. It was commented out for that reason.                                                                 
  - unpair()'s new cleanup sits inside if (m_device && !m_adapterPath.isEmpty()), so when m_adapterPath is empty (connect never succeeded) it still does nothing at all — m_device isn't deleted and m_autoreconnect isn't cleared. Harmless 
    in the FirstPage.qml:13 flow, which clears pairedAddress right after, but the cleanup would be better outside the guard.

piggz:
Logs from daemon look good when starting up

[defaultuser@Quintus ~]$ harbour-amazfishd
2026-09-15 20:31:10.041 : Starting harbour-amazfishd  2.9.0+address.20260915185327.258.g34563fd2
2026-09-15 20:31:10.044 : Failed to load translation for  "en_GB"   "/usr/share/harbour-amazfishd/translations"
2026-09-15 20:31:10.044 : Starting amazfish daemon
2026-09-15 20:31:10.062 : Active voice call changed ""
2026-09-15 20:31:10.067 : Starting notification monitor
2026-09-15 20:31:10.174 : QConnmanEngine: Unable to translate the bearer type of the unknown connection type: "bluetooth"
2026-09-15 20:31:10.305 : void NavigationInterface::connectDaemon() Interface is not valid
2026-09-15 20:31:10.306 : void DeviceInterface::registerDBus()
2026-09-15 20:31:10.306 : Registering service on dbus uk.co.piggz.amazfish
2026-09-15 20:31:10.310 : Succesfully registered to dbus sessionBus
2026-09-15 20:31:10.313 : static AbstractDevice* DeviceFactory::createDevice(const QString&, const QString&) : requested device of type: "Amazfit Balance" "Amazfit Balance"
2026-09-15 20:31:10.314 : AbstractDevice::AbstractDevice(const QString&, QObject*)
2026-09-15 20:31:10.315 : void AbstractDevice::setConnectionState(const QString&) Connection state: "disconnected"
2026-09-15 20:31:10.315 : HuamiDevice::HuamiDevice(const QString&, QObject*)
2026-09-15 20:31:10.315 : ZeppOSDevice::ZeppOSDevice(const QString&, QObject*)
2026-09-15 20:31:10.315 : ZeppOsAuthService::ZeppOsAuthService(ZeppOSDevice*)
2026-09-15 20:31:10.315 : ZeppOsAgpsService::ZeppOsAgpsService(ZeppOSDevice*)
2026-09-15 20:31:10.315 : DBusHRM::DBusHRM(QObject*)
2026-09-15 20:31:10.330 : void DeviceInterface::setupDatabase() DRIVERS:  ("org.kde.kdb.sqlite")
2026-09-15 20:31:10.332 : void DeviceInterface::setupDatabase() : Database is:  "/home/defaultuser/.local/share/harbour-amazfish/harbour-amazfish/amazfish.kexi"
2026-09-15 20:31:10.333 : void DeviceInterface::setupDatabase() : KDbConnection object created.
2026-09-15 20:31:10.333 : void DeviceInterface::setupDatabase() : KDbConnection::connect() OK.
2026-09-15 20:31:10.342 : Upgrading mi_band_activity with INDEX
2026-09-15 20:31:10.349 : ("mi_band_activity", "sports_data", "sports_meta", "info_log", "pai", "hrv", "spo2", "spo2sleep", "temperature", "stress")
2026-09-15 20:31:10.349 : void DeviceInterface::reconnectionTimer()
2026-09-15 20:31:10.349 : void DeviceInterface::reconnectionTimer() Lost connection
2026-09-15 20:31:10.349 : void DeviceInterface::disconnect()
2026-09-15 20:31:10.349 : virtual void AbstractDevice::disconnectFromDevice()
2026-09-15 20:31:10.349 : void AbstractDevice::setConnectionState(const QString&) Connection state: "disconnected"
2026-09-15 20:31:10.349 : QBLEDevice::disconnectFromDevice
2026-09-15 20:31:10.349 : void DeviceInterface::connectToDevice()
2026-09-15 20:31:10.350 : bool DeviceInterface::migrateDataDeviceAddress(const QString&, const QString&) "/org/bluez/hci0/dev_F8_F4_FD_B5_65_04" "F8:F4:FD:B5:65:04"
2026-09-15 20:31:13.730 : void DeviceInterface::connectToDevice(const QString&) : address: "F8:F4:FD:B5:65:04"
2026-09-15 20:31:13.730 : QString DeviceInterface::devicePath(const QString&) "F8:F4:FD:B5:65:04"
2026-09-15 20:31:13.730 : bool DeviceInterface::determineAdapterPath(const QString&) "F8_F4_FD_B5_65_04"
2026-09-15 20:31:13.905 : QMap(("address", QVariant(QString, "1C:9F:4E:72:62:3B"))("alias", QVariant(QString, "halium-algiz"))("itemText", QVariant(QString, "/org/bluez/hci0"))("name", QVariant(QString, ""))("path", QVariant(QString, "/org/bluez/hci0"))) "/org/bluez/hci0/dev_F8_F4_FD_B5_65_04"
2026-09-15 20:31:13.905 : BluezAdapter::BluezAdapter
2026-09-15 20:31:13.905 : BluezAdapter::setAdapterPath: "/org/bluez/hci0"
2026-09-15 20:31:13.926 : void QBLEDevice::setDevicePath(const QString&) "/org/bluez/hci0/dev_F8_F4_FD_B5_65_04"
2026-09-15 20:31:13.934 : void QBLEDevice::setDevicePath(const QString&) "/org/bluez/hci0/dev_F8_F4_FD_B5_65_04" true
2026-09-15 20:31:13.938 : virtual void AbstractDevice::connectToDevice()
2026-09-15 20:31:13.938 : QBLEDevice::disconnectFromDevice
2026-09-15 20:31:13.938 : void AbstractDevice::setConnectionState(const QString&) Connection state: "connecting"
2026-09-15 20:31:13.938 : void DeviceInterface::onConnectionStateChanged() "connecting"
2026-09-15 20:31:13.939 : void HuamiFetcher::reset()
2026-09-15 20:31:13.939 : QBLEDevice::connectToDevice
2026-09-15 20:31:16.336 : void AbstractDevice::setConnectionState(const QString&) Connection state: "disconnected"
2026-09-15 20:31:16.337 : void DeviceInterface::onConnectionStateChanged() "disconnected"
2026-09-15 20:31:16.338 : void HuamiFetcher::reset()
2026-09-15 20:31:23.401 : void AbstractDevice::setConnectionState(const QString&) Connection state: "connected"
2026-09-15 20:31:23.401 : void DeviceInterface::onConnectionStateChanged() "connected"
2026-09-15 20:31:23.401 : void HuamiFetcher::reset()
2026-09-15 20:31:26.086 : initialise() elapsed:  1789500686 starting:  true
2026-09-15 20:31:26.087 : virtual void ZeppOSDevice::initialise()
2026-09-15 20:31:26.087 : void AbstractDevice::parseServices()
2026-09-15 20:31:26.096 : 13 nodes
2026-09-15 20:31:26.112 : virtual QBLEService* ZeppOSDevice::drv_createService(const QString&, const QString&) "00001801-0000-1000-8000-00805f9b34fb"
2026-09-15 20:31:26.115 : virtual QBLEService* HuamiDevice::drv_createService(const QString&, const QString&) "00001801-0000-1000-8000-00805f9b34fb"
2026-09-15 20:31:26.118 : Creating service for:  "00001801-0000-1000-8000-00805f9b34fb" false
2026-09-15 20:31:26.159 : void QBLECharacteristic::introspect() descriptors ("00002902-0000-1000-8000-00805f9b34fb")
2026-09-15 20:31:26.164 : void QBLEService::introspect() characteristics ("00002a05-0000-1000-8000-00805f9b34fb")
2026-09-15 20:31:26.170 : virtual QBLEService* ZeppOSDevice::drv_createService(const QString&, const QString&) "00001800-0000-1000-8000-00805f9b34fb"
2026-09-15 20:31:26.170 : virtual QBLEService* HuamiDevice::drv_createService(const QString&, const QString&) "00001800-0000-1000-8000-00805f9b34fb"
2026-09-15 20:31:26.171 : Creating service for:  "00001800-0000-1000-8000-00805f9b34fb" false
2026-09-15 20:31:26.189 : void QBLECharacteristic::introspect() descriptors ()
2026-09-15 20:31:26.206 : void QBLECharacteristic::introspect() descriptors ()
2026-09-15 20:31:26.221 : void QBLECharacteristic::introspect() descriptors ()
2026-09-15 20:31:26.239 : void QBLECharacteristic::introspect() descriptors ()
2026-09-15 20:31:26.244 : void QBLEService::introspect() characteristics ("00002a00-0000-1000-8000-00805f9b34fb", "00002a01-0000-1000-8000-00805f9b34fb", "00002a04-0000-1000-8000-00805f9b34fb", "00002aa6-0000-1000-8000-00805f9b34fb")
2026-09-15 20:31:26.250 : virtual QBLEService* ZeppOSDevice::drv_createService(const QString&, const QString&) "00001811-0000-1000-8000-00805f9b34fb"
2026-09-15 20:31:26.251 : virtual QBLEService* HuamiDevice::drv_createService(const QString&, const QString&) "00001811-0000-1000-8000-00805f9b34fb"
2026-09-15 20:31:26.269 : void QBLECharacteristic::introspect() descriptors ()
2026-09-15 20:31:26.274 : void QBLEService::introspect() characteristics ("00002a46-0000-1000-8000-00805f9b34fb")
2026-09-15 20:31:26.276 : AlertNotificationService::AlertNotificationService(const QString&, QObject*, uint8_t)
2026-09-15 20:31:26.277 : Creating service for:  "00001811-0000-1000-8000-00805f9b34fb" true
2026-09-15 20:31:26.282 : virtual QBLEService* ZeppOSDevice::drv_createService(const QString&, const QString&) "0000180a-0000-1000-8000-00805f9b34fb"
2026-09-15 20:31:26.282 : virtual QBLEService* HuamiDevice::drv_createService(const QString&, const QString&) "0000180a-0000-1000-8000-00805f9b34fb"
2026-09-15 20:31:26.301 : void QBLECharacteristic::introspect() descriptors ()
2026-09-15 20:31:26.318 : void QBLECharacteristic::introspect() descriptors ()
2026-09-15 20:31:26.335 : void QBLECharacteristic::introspect() descriptors ()
2026-09-15 20:31:26.353 : void QBLECharacteristic::introspect() descriptors ()
2026-09-15 20:31:26.369 : void QBLECharacteristic::introspect() descriptors ()
2026-09-15 20:31:26.375 : void QBLEService::introspect() characteristics ("00002a23-0000-1000-8000-00805f9b34fb", "00002a25-0000-1000-8000-00805f9b34fb", "00002a27-0000-1000-8000-00805f9b34fb", "00002a28-0000-1000-8000-00805f9b34fb", "00002a50-0000-1000-8000-00805f9b34fb")
2026-09-15 20:31:26.377 : DeviceInfoService::DeviceInfoService(const QString&, QObject*)
2026-09-15 20:31:26.378 : Creating service for:  "0000180a-0000-1000-8000-00805f9b34fb" true
2026-09-15 20:31:26.384 : virtual QBLEService* ZeppOSDevice::drv_createService(const QString&, const QString&) "0000fee0-0000-1000-8000-00805f9b34fb"
2026-09-15 20:31:26.384 : virtual QBLEService* HuamiDevice::drv_createService(const QString&, const QString&) "0000fee0-0000-1000-8000-00805f9b34fb"
2026-09-15 20:31:26.413 : void QBLECharacteristic::introspect() descriptors ("00002902-0000-1000-8000-00805f9b34fb")
2026-09-15 20:31:26.439 : void QBLECharacteristic::introspect() descriptors ("00002902-0000-1000-8000-00805f9b34fb")
2026-09-15 20:31:26.466 : void QBLECharacteristic::introspect() descriptors ("00002902-0000-1000-8000-00805f9b34fb")
2026-09-15 20:31:26.493 : void QBLECharacteristic::introspect() descriptors ("00002902-0000-1000-8000-00805f9b34fb")
2026-09-15 20:31:26.519 : void QBLECharacteristic::introspect() descriptors ("00002902-0000-1000-8000-00805f9b34fb")
2026-09-15 20:31:26.545 : void QBLECharacteristic::introspect() descriptors ("00002902-0000-1000-8000-00805f9b34fb")
2026-09-15 20:31:26.573 : void QBLECharacteristic::introspect() descriptors ("00002902-0000-1000-8000-00805f9b34fb")
2026-09-15 20:31:26.598 : void QBLECharacteristic::introspect() descriptors ("00002902-0000-1000-8000-00805f9b34fb")

jmlich:
Three things left, one of them a consequence of the bump.

  1. setDevicePath() leaks a QDBusInterface on every retry

  qble/qbledevice.cpp:7-20 (at 16102222) still overwrites m_deviceInterface without deleting the previous one:

  void QBLEDevice::setDevicePath(const QString& path)
  {
      m_devicePath = path;
      m_deviceInterface = new QDBusInterface("org.bluez", m_devicePath, "org.bluez.Device1", ...);
   
  That used to be harmless because it was called once per connect from the UI. Moving the retry loop into DeviceInterface changed that: reconnectionTimer() → connectToDevice() → connectToDevice(address) calls
  m_device->setDevicePath(m_deviceAddress) (deviceinterface.cpp:107) on every tick where the BlueZ device object still exists — which is the normal case for a watch that is merely switched off or out of range, since the dev_… node
  persists for known devices. The old AbstractDevice::reconnectionTimer() went straight to QBLEDevice::connectToDevice() and reused the existing interface.

  So it's one leaked QDBusInterface per minute, indefinitely. (The repeated QDBusConnection::connect() in the same function is fine — Qt dedupes identical service/path/signature/receiver/slot hooks — so this is only the leak, not
  duplicate signal delivery.)

  Cheapest fix is on the amazfish side, since devicePath() is already exposed (qbledevice.h:24):
 
  if (m_device->devicePath() != m_deviceAddress) {
      m_device->setDevicePath(m_deviceAddress);
  }   

  Deleting the old interface inside setDevicePath() would be the more complete fix, but it needs the matching QDBusConnection::disconnect() for the old path too.

  2. AdapterModel still leaks its m_objectManager
  
  qble/adaptermodel.cpp:14 — the bump gave BluezAdapter a destructor but AdapterModel still has none (adaptermodel.h:11), and it's the one constructed fresh inside determineAdapterPath() on every tick. Same one-per-minute rate. A
  ~AdapterModel() mirroring the new ~BluezAdapter(), or just passing this as the interface's parent, closes it.

  3. The toast every 60 seconds (unchanged)

  deviceinterface.cpp:111 still emits message(tr("Device is not yet available")) on each failed resolution, and the timer drives it once a minute whenever the device isn't in BlueZ at all. Worth emitting only on the first failure of a
  run, or suppressing it entirely on the timer path.
  
  Housekeeping
 
  - git status in the superproject shows M qble: the submodule working tree is still checked out at 35d1a8bf while the superproject now records 16102222. Run git submodule update — otherwise the next commit that stages qble will silently
    revert the bump.
  - devicePath's declaration and definition still disagree on spacing (const QString &address in deviceinterface.h:125, const QString& address in .cpp:792). d581b697 aligned migrateDataDeviceAddress on &x but left this one split.

jmlich:
● Reviewed fbc51353 + 66ff52da (and the two new qble commits). The submodule is in sync now, so that housekeeping item is closed.

  1. The setDevicePath leak guard misses the case that actually leaks
 
  qble/qbledevice.cpp:21-27:
   
  if (m_deviceInterface && (path != m_devicePath)) {
      delete m_deviceInterface;
      QDBusConnection::systemBus().disconnect(...m_devicePath...);
  }

  m_devicePath = path;
  m_deviceInterface = new QDBusInterface(...);

  When the path is unchanged, the old interface is not deleted — but the next line overwrites the pointer anyway. So it still leaks. And unchanged is exactly the retry case: same watch, same adapter, DeviceInterface::connectToDevice()
  calls setDevicePath with an identical string every 60 seconds. The new guard helps when the path changes, which is the case that never happens in this flow.

  An early return fixes it and also skips the redundant connect() re-registration:
   
  if (m_deviceInterface && path == m_devicePath) {
      return; // same device, nothing to rebuild
  }
   
  if (m_deviceInterface) {
      QDBusConnection::systemBus().disconnect("org.bluez", m_devicePath, "org.freedesktop.DBus.Properties", "PropertiesChanged", argumentMatch, QString(), this, SLOT(onPropertiesChangedInt(QString, QVariantMap, QStringList)));
      delete m_deviceInterface;
      m_deviceInterface = nullptr;
  }

  Two smaller things in the same block: disconnect before the delete (the current order uses m_devicePath after freeing the interface — safe today, since only the path string is read, but it is an easy trap), and null the pointer after
  deleting.
 
  2. BluezAdapter::setAdapterPath has the same shape
  
  qble/bluezadapter.cpp:20 assigns a new QDBusInterface with no delete of the previous one. No caller calls it twice on one object today (the two internal calls at :27 and :39 are guarded by !m_deviceInterface), so it does not leak yet.
  Same trap though — worth closing in the same commit.
 
  3. AdapterModel destructor: good
  
  qble/adaptermodel.cpp:34-37 closes that leak. What remains is cost, not a bug: determineAdapterPath still builds a fresh AdapterModel per tick, and Adam's log measured that GetManagedObjects at ~175 ms of blocked main thread. If you
  ever want to trim the failing-retry loop, caching m_adapterPath and rescanning only when the cached path stops resolving would drop it to near zero. Optional.
 
  4. The toast guard also silences manual retries
 
  deviceinterface.cpp:112-115 — the flag goes false on the first failure and back to true only at "authenticated" (:721). The once-a-minute toast is gone, which was the point. But so is feedback on a user-initiated connect: watch off,
  user taps Connect → toast; taps again → nothing. The UI cannot tell "trying" from "ignored".
 
  Both paths go through the no-arg connectToDevice(), so the split has to happen there:
 
  void DeviceInterface::connectToDevice()   // Q_INVOKABLE, from the UI
  {
      m_allowDeviceNotAvailableMessage = true;   // user asked, so always answer
      autoConnectToDevice();
  }
   
  with reconnectionTimer() (:157) calling autoConnectToDevice() — the current body — instead. The timer then never speaks, the user always gets an answer.

  Nit: the name describes the mechanism. m_deviceNotAvailableReported, inverted, reads closer to what it holds. Take it or leave it.

  5. Pre-existing, but the new line sits on top of it

  deviceinterface.cpp:723-725:
  
  m_allowDeviceNotAvailableMessage = true;

  m_device->setDatabase(dbConnection());
  if (m_device) {

  m_device is used one line before it is checked. Not from this branch, and it cannot be null at "authenticated" in practice — but since you are editing these exact lines, moving setDatabase inside the if is free.

  Still open from the log round

  - The migration's return value is never logged, so a failed migration and a successful one produce identical logs. One qDebug at deviceinterface.cpp:133 makes it verifiable.
  - The transient "disconnected" during an in-flight connect (7 s in Adam's log) is still a window where the retry timer can tear down a connection that was about to succeed.
 
  Everything else from rounds 1-5 is fixed. Findings 1 and 4 are the ones worth another commit; 2, 3, and 5 are tidy-ups.

piggz:
Ready for another round!

jmlich:
Fixed and confirmed

  - The setDevicePath leak is genuinely closed now. qble/qbledevice.cpp:21-33 — same path is a full no-op: no new interface, no duplicate connect(). That is the retry case, which is what was leaking.
  - BluezAdapter::setAdapterPath deletes before reassigning (qble/bluezadapter.cpp:22).
  - Migration result is logged (deviceinterface.cpp:1219), so the next log tells you whether it worked.
  - userInitiated plumbing works. Timer passes false (:159), UI passes true (ui/src/daemoninterface.cpp:113). Per-call now, not sticky — cleaner than the old flag.
  - Submodule in sync at 454fc3f3.

  1. New null deref in setDevicePath

  qble/qbledevice.cpp:31 — the qDebug moved outside the guarded block:
 
  if (path != m_devicePath) {
      ...
      m_deviceInterface = new QDBusInterface(...);
      ...
  }
  
  qDebug() << Q_FUNC_INFO << path <<  m_deviceInterface->isValid();
  
  On the first call with an empty path, path == m_devicePath (both empty), the block is skipped, and m_deviceInterface is still null. Both current callers guard on non-empty (deviceinterface.cpp:107 and :167), so it cannot fire today —
  but it is a crash added to the function being hardened. Move the qDebug inside the if.

  2. Same hole in setAdapterPath, and it disables a recovery path

  qble/bluezadapter.cpp:18-20:
  
  if (path == m_adapterPath) {
      return;
  }
  
  startDiscovery (:31-35) and stopDiscovery (:39-43) both do:
  
  if (!m_deviceInterface && !m_adapterPath.isEmpty()) {
      setAdapterPath(m_adapterPath);
  }
  
  QDBusMessage reply = m_deviceInterface->call("StartDiscovery");
 
  That call now returns without building the interface, and the next line dereferences null. Rebuilding a missing interface is the only reason those two lines exist, and the new guard defeats them. Unreachable today (m_adapterPath is
  only ever set inside setAdapterPath), but it reads as handled and is not.

  Both this and finding 1 fix the same way — test the pointer too:

  if (path == m_adapterPath && m_deviceInterface) {
      return;
  }
 3. The new !m_device return also swallows the state change
  
  deviceinterface.cpp:722-724 sits above emit connectionStateChanged(); at :760, so a null device now drops the signal the UI listens on.
                                                                                                                                                                                                                                             
  It cannot fire today: the slot is only connected to m_device's own signal, direct connection, and m_device is deleted only in pair() (:176) and unpair() (:216), which disconnects it. So this is defensive. But if it ever did fire, the
  UI would sit on a stale state instead of crashing — a worse failure, and a silent one. Keep the emit reachable:

  if (m_device) {
      if (connectionState() == "authenticated") {
          ...
      } else {
          ...
      }
  }   
          
  emit connectionStateChanged();

  4. Drop the default argument
  
  deviceinterface.h:50 — Q_INVOKABLE void connectToDevice(bool userInitiated = true);
  
  This object is exported with ExportAllInvokables (:1354), so the default argument lands in the meta-object as a second, cloned entry. That is the same overload shape you removed in eba6467c. Nothing calls the no-arg form any more — the
  timer passes false, the UI passes true over the bus, and QML calls DaemonInterface::connectToDevice() in the UI binary, not the daemon's. So making it required costs nothing and leaves exactly one exported signature:

  Q_INVOKABLE void connectToDevice(bool userInitiated);

  Either way the bus signature changed from connectToDevice() to connectToDevice(b). In-tree that is consistent, but an old UI against a new daemon gets "no such method". Worth a line in the release note if UI and daemon can ship apart.

  5. Leftover

  deviceinterface.cpp:743 is still if (m_device && m_device->supportsFeature(...)) while :752 was cleaned to drop the check. Same function, now inconsistent. Trivial.

  Findings 1 and 2 are the ones to fix; 3, 4 and 5 are small. The "disconnected" race from the log round is still open.

jmlich:
LLM looks happy now
All five items from the last round are fixed, and the fixes are correct:

  - qbledevice.cpp:18 — empty path returns early. Combined with m_devicePath only ever being assigned inside the block that creates the interface, the trailing m_deviceInterface->isValid() can no longer see a null pointer. Closed
    properly.
  - bluezadapter.cpp:18,31,44 — the dead self-healing calls are gone and replaced with a plain guard + log. Nothing in this repo calls qble's startDiscovery/stopDiscovery (the pairing page uses BluezQt's adapter), so no behaviour change.
  - deviceinterface.cpp:722-757 — emit connectionStateChanged() is now always reached; the redundant m_device && is gone.
  - deviceinterface.h:50 — default argument removed, so one exported D-Bus signature. No caller was left needing the no-arg form.

  Nothing new in this diff. The two open items are both pre-existing behaviour, not regressions, and neither blocks merge:

  1. The transient "disconnected" during an in-flight connect (7 s in the log) is still a window where the 60 s timer can tear down a connection that was about to succeed.
  2. determineAdapterPath still does a ~200 ms blocking adapter scan per retry.

  From my side the branch looks ready. Worth one more run of Adam's startup log to confirm the migration line now prints true.

